PR #8, opened 2021-03-30
add parameters to make-deal

acruikshank:
We're still working on getting deals working, but this PR introduces some parameters to the deal that are helpful for coordination.

### DIFF
--- a/commands/deal.go
+++ b/commands/deal.go
@@ -2,6 +2,7 @@ package commands
 
 import (
 	"context"
+	"crypto/rand"
 	"fmt"
 	"github.com/c2h5oh/datasize"
 	"github.com/filecoin-project/dealbot/lotus"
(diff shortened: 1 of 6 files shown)


commands/deal.go
```diff
@@ -1,34 +1,136 @@
...
+	Name:  "MakeDeal",
 	Usage: "Make storage deals with provided miners.",
 	Flags: []cli.Flag{
-		// TODO: parameters?
+		&cli.PathFlag{
+			Name:     "data-dir",
+			Usage:    "writable directory used to transfer data to node",
+			Aliases:  []string{"d"},
+			EnvVars:  []string{"DEALBOT_DATA_DIRECTORY"},
+			Required: true,
```
willscott, line 29: I think we want 2 parameters for this:
where that directory is from the perspective of the dealbot, and where it is from the perspective of the lotus node.

willscott (approved):

willscott:
i see a few things i want, but i'll defer them to issues for follow up PRs

PR #11, opened 2021-03-31
add retrieval-deal cmd

nonsense:
Initial `retrieval-deal` cmd and slight refactor of function names to allow for it.

TODO:
- [ ] fix issues with using default wallet from lotus node


commands/retrieval_deal.go
```diff
@@ -0,0 +1,139 @@
...
+
+	rpath, err := ioutil.TempDir("", "dealbot-retrieve-test-")
+	if err != nil {
+		panic(err)
+	}
+	defer os.RemoveAll(rpath)
+
+	caddr, err := client.WalletDefaultAddress(ctx)
+	if err != nil {
+		return err
```
nonsense, line 111: Currently we fail here, as we can't just use the default wallet of Lotus via an external process.

commands/retrieval_deal.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	v, err := node.Version(context.Background())
+	if err != nil {
+		return err
+	}
+
+	log.Infof("remote version: %s", v.Version)
```
acruikshank, line 59: This is a good idea. I'll add it to make deal too.

commands/retrieval_deal.go
acruikshank: I parameterized this in make deal: https://github.com/filecoin-project/dealbot/blob/fb6f863634aae55eeddbeef4743101a39715e4f0/commands/deal.go#L40

commands/retrieval_deal.go
```diff
@@ -0,0 +1,139 @@
...
+
+	log.Info("got offer")
+
+	//spew.Dump(offer)
+
+	//if len(offers) < 1 {
+	//return errors.New("no offers")
+	//}
+
+	rpath, err := ioutil.TempDir("", "dealbot-retrieve-test-")
```
acruikshank, line 103: @willscott suggested that this script and the lotus node might not be able to see the filesystem from the same perspective. This should probably be parameterized. See: https://github.com/filecoin-project/dealbot/blob/fb6f863634aae55eeddbeef4743101a39715e4f0/commands/deal.go#L28

acruikshank (commented):
It would be nice to get the parameters for this and the make-deal command more consistent. This draft is looking good though and I'll be excited to hear it's working.

nonsense:
This works if the `dealbot` is deployed alongside the `lotus node` so that it has access to the `repo` and the `wallet`.

It needs more work with respect to having permissions to trigger retrieval deals with a given wallet over the API of the lotus node.

Issue #14, opened 2021-04-01
Remove cache layer and additional interfaces in front of api.FullNode


nonsense:
At the moment we have a lot of indirection that seems unnecessary, such as `CacheCtxStore`, `APIWrapper`, `API` interface,    `APIOpener` and `APICloser`. None of these seem to bring any more value that just directly using the Lotus API, namely `api.FullNode`.

Every time we want to use a new API endpoint from `api.FullNode`, we have to edit a bunch of structs to add it to our interfaces on `dealbot` side, which seems redundant.

PR #16, opened 2021-04-01
remove redundant interfaces

nonsense:
Fixes: https://github.com/filecoin-project/dealbot/issues/14

Subsumes: https://github.com/filecoin-project/dealbot/pull/15

---

As discussed yesterday in standup and in the #dealbot Slack channel, I went ahead and removed the indirection that currently doesn't seem to bring much value to the `dealbot`.


acruikshank (approved):
Yeah, I was already getting a little frustrated with this. I got started on some test infrastructure that I haven't pushed up yet, that was using the interface to fake the lotus API. Without an interface this won't be possible. We might want to consider bringing back just that part of it, but this looks good for now.

nonsense:
We could probably mock directly on the `api.FullNode` level/interface if we want to add tests on that border.

PR #20, opened 2021-04-05
initial dealbot daemon

nonsense:
This PR is adding a `daemon` command, with the idea that `dealbot daemon` would run a single HTTP server to listen for commands from the `dealbot controller`.

`dealbot daemon` should accept `tasks` from the `dealbot controller` and report back results (directly writing to stdout, and probably shorter response to the `dealbot controller`.

---

This is still in heavy WIP.

### DIFF
--- a/commands/setup.go
+++ b/commands/setup.go
@@ -51,11 +51,11 @@ func setupLotusAPI(cctx *cli.Context) (*lotus.APIOpener, lotus.APICloser, error)
 
 var CommonFlags []cli.Flag = []cli.Flag{
 	&cli.PathFlag{
-		Name:     "data-dir",
-		Usage:    "writable directory used to transfer data to node",
-		Aliases:  []string{"d"},
-		EnvVars:  []string{"DEALBOT_DATA_DIRECTORY"},
-		Required: true,
+		Name:    "data-dir",
+		Usage:   "writable directory used to transfer data to node",
+		Aliases: []string{"d"},
+		EnvVars: []string{"DEALBOT_DATA_DIRECTORY"},
+		//Required: true,
 	},
 	&cli.PathFlag{
 		Name:    "node-data-dir",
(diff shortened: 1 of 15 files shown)


willscott (commented):
The structure of running a daemon looks good.

My guess is that it'll be easier for the `controller` to be a public http endpoint that we configure the individual dealbots to request tasks from, rather than each dealbot running an http server that the controller makes requests to.

I suspect we will find it more complicated from an ops perspective to expose the http server of the individual deal bots to a central controller than the other way around.

commands/setup.go
```diff
@@ -70,11 +70,11 @@ var CommonFlags []cli.Flag = []cli.Flag{
...
-		Name:     "miner",
-		Usage:    "address of miner to make deal with",
-		Aliases:  []string{"m"},
-		EnvVars:  []string{"DEALBOT_MINER_ADDRESS"},
-		Required: true,
+		Name:    "miner",
+		Usage:   "address of miner to make deal with",
+		Aliases: []string{"m"},
+		EnvVars: []string{"DEALBOT_MINER_ADDRESS"},
+		//Required: true,
```
acruikshank, line 77: This parameter is going to be needed for storage deals. Why is not required?

acruikshank (commented):
I'm with @willscott. This daemon code looks great, but it makes more sense for the clients to pull tasks and push results.

commands/setup.go
nonsense: I removed it, as `required` because I am not sure it should be specified with the `daemon` command. The `dealbot daemon` runs alongside a `lotus-miner`, and makes deals with MANY miners, whereas this option is for the CLI to specify ONE miner to make a deal with.

nonsense:
> My guess is that it'll be easier for the controller to be a public http endpoint that we configure the individual dealbots to request tasks from, rather than each dealbot running an http server that the controller makes requests to. I suspect we will find it more complicated from an ops perspective to expose the http server of the individual deal bots to a central controller than the other way around.

Agreed!

willscott (approved):

PR #21, opened 2021-04-06
Refactor commands to task and add JSON parsing

acruikshank:
### Motivation

This is an intermediate step between a dealbot run entirely from the command line and a distributed dealbot where clients pick up tasks from a centralized coordinator. Specifically this PR introduces the tasks and a `coordinate` command (which might not live past this PR)` to read tasks from a JSON file and execute them in series.

### What's in this PR

1. Refactor storage-deal and retrieval-deal logic into two corresponding tasks and task runners.
2. Further refine common parameters for commands.
3. Refactor common setup logic for all commands into `setupCLIClient`.
4. Add JSON parsing for tasks.
5. Introduce `coordinate` command that parses a JSON file and then runs the tasks sequentially.

### DIFF
--- a/commands/coordinate.go
+++ b/commands/coordinate.go
@@ -1,0 +1,61 @@
+package commands
+
+import (
+	"context"
+	"net/http"
+	"time"
+
+	"github.com/filecoin-project/dealbot/config"
+	"github.com/filecoin-project/dealbot/daemon"
+	"github.com/urfave/cli/v2"
+)
+
+var CoordinateCmd = &cli.Command{
+	Name:   "coordinate",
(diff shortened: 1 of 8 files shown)


commands/coordinate.go
```diff
@@ -0,0 +1,90 @@
...
+		}
+		if err != nil {
+			log.Error(err)
+		}
+	}
+
+	return nil
+}
+
+func parseTasks() ([]interface{}, error) {
```
nonsense, line 58: I think using something like `github.com/mitchellh/mapstructure` would be a great fit here, although this is fine for now.

nonsense (approved):

PR #22, opened 2021-04-06
introduce `dealbot controller`

nonsense:
This PR is adding the `dealbot controller` command, which starts the controller component of the dealbot project, namely the service that keeps track of all `tasks` and that individual dealbots fetch work from and later update with the status of individual tasks.

At the moment the state of all tasks is in-memory until we decide we are happy with the model - later we should introduce persistence.

`dealbot controller` should feature RPC endpoints that allow for:
1. acquiring a task from the currently `available` tasks
2. updating a task with its status as the given `dealbot` sees it.


willscott (approved):

controller/controller.go
```diff
@@ -0,0 +1,82 @@
...
+	// Set a unique request ID.
+	r.Use(func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			r.Header.Set("X-Request-ID", uuid.New()[:8])
+			next.ServeHTTP(w, r)
+		})
+	})
+
+	r.HandleFunc("/tasks", srv.getTasksHandler).Methods("GET")
+	//r.HandleFunc("/task", srv.getTaskHandler).Methods("GET")
```
acruikshank, line 39: This should be a POST once the server starts serving tasks from a queue and tracking their status.

acruikshank (commented):
Like all the naming changes. Server looks like a good start.

PR #23, opened 2021-04-07
Add client

acruikshank:
### Motivation

We want dealbot to be able to run on a remote machine with centralized control. This PR introduces a `client` command (actually replaces the `coordinator` command) that requests tasks from the controller and pushes status updates to it.

### Proposed Changes

1. Change `coordinator` command to `client`.
2. Have `client` take a url to the controller and parse tasks from an HTTP request rather than stdin.
3. Add a new controller endpoint to accept and log a status message from the client.
4. Modify the client logger to post status updates to the server.


willscott:
would this also replace `daemon` mode?

acruikshank:
@willscott It might replace daemon mode or it might need to be integrated into it. We should discuss that.

nonsense (approved):

nonsense:
> would this also replace daemon mode?

As discussed yesterday, this is not replacing the daemon mode, and the daemon mode currently handles the HTTP server as well as the Engine that queries the Controller for Tasks.

PR #24, opened 2021-04-07
initial state for tasks; function to acquire task

nonsense:
This PR is:
1. Modifying the `State` for the controller (still in-memory) and introducing functionality to acquire a task by a given dealbot and mark it as `in-progress`.
2. Adding `controller client` type to be used by the `daemon` to communicate easily with the `controller`.
3. Refactoring the `lotus client` interface from the `dealbot` to its local `lotus node`.
4. Initial handling of `retrieve` and `store` tasks by the `daemon`.
5. Adding `--configpath` flag to provide `toml` configuration on startup - easier than handling HOME folders, etc.


config/loader.go
```diff
@@ -10,7 +10,7 @@ const (
 	EnvDealbotHome = "DEALBOT_HOME"
 
 	// DefaultDaemonListenAddr is a host:port value, where we set up an HTTP endpoint.
-	DefaultDaemonListenAddr = "localhost:8765"
+	DefaultDaemonListenAddr = "localhost:0"
```
nonsense, line 13: Changing to :0 for now, so that I can easily run a few of these locally while testing.

daemon/daemon.go
```diff
@@ -27,6 +28,9 @@ func New(cfg *config.EnvConfig) (srv *Daemon, err error) {
 
 	r := mux.NewRouter().StrictSlash(true)
 
+	e := engine.New(cfg)
```
nonsense, line 31: Will need to pass this to the HTTP server, so that all the `daemon handlers` have access to the `engine` and can communicate useful information via their HTTP interface.

engine/engine.go
```diff
@@ -0,0 +1,85 @@
...
+)
+
+var log = logging.Logger("engine")
+
+type Engine struct {
+	client *client.Client
+}
+
+func New(cfg *config.EnvConfig) *Engine {
+	workers := 2
```
nonsense, line 22: How many concurrent workers we want to have per dealbot to work on tasks - for now I have hardcoded 2, but we can easily make this configurable. I remember a limit of GraphSync of 6, so we might have issues running >6 concurrent retrieval deals - TBD.

controller/client/client.go
```diff
@@ -0,0 +1,97 @@
...
+	client   *http.Client
+	cfg      *config.EnvConfig
+	endpoint string
+}
+
+// New initializes a new API client
+func New(cfg *config.EnvConfig) *Client {
+	endpoint := cfg.Client.Endpoint
+
+	log.Infow("dealbot controller client initialized", "addr", endpoint)
```
willscott, line 31: aim for key/value pairs on these structured logs, iirc

engine/engine.go
willscott: seems fine

controller/client/client.go
nonsense: I am not sure I understand what you mean? This is already a structured log - i.e. a description + key/value pair.

willscott: sounds good. i think i missed the 'w' - i've seen others where the description isn't implicitly tagged, so you need to do `"msg", "...", "addr", endpoint`

nonsense:
Marking this as ready for review, as it is consistently better than what we have on `master`.

commands/controller.go
```diff
@@ -11,17 +11,26 @@ import (
...
 var ControllerCmd = &cli.Command{
-	Name:   "controller",
-	Usage:  "Start the dealbot controller",
+	Name:  "controller",
+	Usage: "Start the dealbot controller",
+	Flags: []cli.Flag{
+		&cli.StringFlag{
+			Name:  "configpath",
+			Usage: "path to config toml file",
+			Value: ".env.toml",
```
willscott, line 20:
```suggestion
			Value: ".dealbot-controller.env.toml",
```

commands/daemon.go
```diff
@@ -11,17 +11,26 @@ import (
...
 var DaemonCmd = &cli.Command{
-	Name:   "daemon",
-	Usage:  "Start a dealbot daemon, accepting tasks over RPC",
+	Name:  "daemon",
+	Usage: "Start a dealbot daemon, accepting tasks over RPC",
+	Flags: []cli.Flag{
+		&cli.StringFlag{
+			Name:  "configpath",
+			Usage: "path to config toml file",
+			Value: ".env.toml",
```
willscott, line 20:
```suggestion
			Value: ".dealbot.env.toml",
```

daemon/daemon.go
```diff
@@ -27,6 +28,12 @@ func New(cfg *config.EnvConfig) (srv *Daemon, err error) {
 
 	r := mux.NewRouter().StrictSlash(true)
 
+	e, err := engine.New(cfg)
```
willscott, line 31: defer a Close to it to clean up after there's a signal to stop?

dealbot.toml
```diff
@@ -0,0 +1,12 @@
+[controller]
+listen = "localhost:8764"
+
+[daemon]
+listen        = "localhost:0"
+data_dir      = "/Users/nonsense/code/src/github.com/filecoin-project/dealbot"
+node_data_dir = "/Users/nonsense/code/src/github.com/filecoin-project/dealbot"
+api           = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJBbGxvdyI6WyJyZWFkIiwid3JpdGUiLCJzaWduIiwiYWRtaW4iXX0.Q9gom7KbWTT8XRFMVMC6Ds5uIBk7KKb8xhdVlEgRyk4:/ip4/127.0.0.1/tcp/1234/http"
```
willscott, line 8: maybe don't commit? 😅

daemon/daemon.go
nonsense: Added `engine` to `daemon`, so now it is handled in `srv.Shutdown`

daemon/daemon.go
```diff
@@ -20,13 +21,19 @@ type Daemon struct {
...
 	doneCh chan struct{}
+	e      *engine.Engine
 }
 
 func New(cfg *config.EnvConfig) (srv *Daemon, err error) {
 	srv = new(Daemon)
 
 	r := mux.NewRouter().StrictSlash(true)
 
+	srv.e, err = engine.New(cfg)
```
acruikshank, line 32: Do we need Daemon? It appears to be creating an engine and starting an HTTP server which doesn't currently serve anything. I don't believe we should be starting a server on the client. Unless there is a reason we need a server on both sides, we could just construct the engine from the daemon command.

controller/state.go
```diff
@@ -1,29 +1,67 @@
...
+}
+
+func (s *state) Update(req *client.UpdateTaskRequest) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	for _, t := range s.tasks {
+		if t.UUID == req.UUID {
+			if t.Status == tasks.Available {
+				log.Infow("state update", "uuid", t.UUID, "status", req.Status, "worked_by", req.WorkedBy)
```
nonsense, line 32: This needs a lot more improvement. This is just the basic `update` from `Available` to `InProgress` - we will need something much more elaborate to properly track Tasks and updates.

acruikshank (approved):
This looks good. I don't understand why we have a daemon and an engine. I feel pretty strongly we should not be starting a server on the client side. We should be passing a context down to the tasks so they can exit gracefully. Other than that, everything looks great.

daemon/daemon.go
nonsense: We are starting an HTTP server on the `dealbot daemon` just so that we serve metrics in the future, and currently we block on the HTTP server listen I believe. You are right, that it doesn't do anything at the moment, but I think it will come handy very soon - for metrics, health, etc.

engine/engine.go
```diff
@@ -0,0 +1,137 @@
...
+func (e *Engine) Close() {
+	e.closer()
+}
+
+func (e *Engine) worker(n int) {
+	log.Infow("engine worker started", "worker_id", n)
+
+	for {
+		func() {
+			time.Sleep(5 * time.Second)
```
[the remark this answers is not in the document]
nonsense: This serves the purpose of a small delay for when a bot queries the controller, in order not to fire hundreds of requests if there are no available jobs/tasks. At the moment we have 4 tasks hard-coded in the controller, and when they are all taken, we don't want the dealbots to end in a busy loop here.

nonsense:
> I don't understand why we have a daemon and an engine. I feel pretty strongly we should not be starting a server on the client side.

The `daemon` is the process that starts:
1. HTTP server for /metrics /health and maybe other auxiliary HTTP endpoints per dealbot - if we decide we don't need them, we should remove it, but I believe we will need them as this process will live in a Kubernetes pod?
2. Engine is the actor that handles the tasks from the controller.

> We should be passing a context down to the tasks so they can exit gracefully.

Totally agree, I haven't done that diligently yet.

daemon/daemon.go
acruikshank: The metrics could be pushed to the controller. We should keep our communication patterns standard, otherwise we're just going to complicate deployment.

engine/engine.go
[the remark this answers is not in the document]
nonsense: Added a note, although this will disappear as soon as we have a more sane Engine / Task Manager.

PR #30, opened 2021-04-12
try out testscript

nonsense:
An initial attempt to add an integration test to the dealbot.


cmd/devnet/localnet.json
```diff
@@ -0,0 +1,299 @@
+{
```
willscott, line 1: add to .gitignore

willscott (commented):
this structure looks good so far

testdata/scripts/first.txt
```diff
@@ -0,0 +1,10 @@
+exec devnet &
+
+exec sleep 20
+
+env TOKEN=`cat ~/.devLotus/token`
+env API=`cat ~/.devLotus/api`
+env DATADIR=/tmp
+env MINER=t01000
+
+exec dealbot --api $TOKEN:$API storage-deal --data-dir $DATADIR --miner $MINER
```
mvdan, line 10: you should add dealbot as a TestMain command, like so: https://github.com/burrowers/garble/blob/3afc993266c3f331bc116eeea0fd224a266c64de/main_test.go#L33

Then here you can `exec dealbot` or just `dealbot`, and it will run the binary magically without you having to `go install` it into the global PATH previously.

testdata/scripts/first.txt
```diff
@@ -0,0 +1,10 @@
+exec devnet &
+
+exec sleep 20
+
+env TOKEN=`cat ~/.devLotus/token`
+env API=`cat ~/.devLotus/api`
```
mvdan, line 6: command substitutions don't work here, as this is not a full shell. Assuming these are generated and non-static, one way to fix this would be to define a command like `read-env TOKEN .devLotus/token`, using Cmds in https://pkg.go.dev/github.com/rogpeppe/go-internal/testscript#Params and a combination of https://pkg.go.dev/github.com/rogpeppe/go-internal/testscript#TestScript.ReadFile and https://pkg.go.dev/github.com/rogpeppe/go-internal/testscript#TestScript.Setenv.

testdata/scripts/first.txt
```diff
@@ -0,0 +1,10 @@
+exec devnet &
```
mvdan, line 1: just like dealbot, add this to TestMain

nonsense: Right, then we only have to assume that `lotus` is available in the PATH, compiled for 2k sectors.

mvdan: :+1: you could also do something like `[!exec:lotus] skip 'test needs lotus to run'` if you would rather have the tests be skipped rather than fail when lotus isn't installed.

nonsense:
Closing in favour of https://github.com/filecoin-project/dealbot/pull/38

PR #32, opened 2021-04-13
don't use two uuid libraries at once

mvdan:
(see commit message)


mvdan:
Err, I hadn't noticed that the API names are the same, but the types are not.

nonsense (approved):

PR #37, opened 2021-04-13
allow override of config file settings from env for docker deploy

willscott:
currently a bunch of options like the bind interface the controller listens to are specified exclusively through the config file.
That's annoying to set up in a docker environment. This allows any of those config settings to also be taken from environment variables.


mvdan (commented):
I might be missing something here, but I thought CLI frameworks like urfave/cli already supported this out of the box: https://github.com/urfave/cli/blob/master/docs/v2/manual.md#values-from-the-environment

mvdan:
Ah, hm, I see this is the config and not related to urfave or the flags.

I'm not too worried about the added code per se, but since we're already using a CLI framework it feels a bit weird to roll our own reflection :) You could also consider running https://github.com/kelseyhightower/envconfig after loading from the TOML file, since the overriding logic should work just the same.

nonsense:
I think what @mvdan is saying is correct - we could just add an additional config next to the CLI flags, so that they are taken from the environment variables. I am not sure what the precedence is there if multiple are set though.

mvdan:
I'm pretty sure that urfave and viper both have the same precedence: config, flag, env var, where the last wins. I agree we should just rely on urfave and make all settings consistent.

willscott:
if we want to have everything specified as flags, i'm happy with that.
@nonsense - I think you added the config file structure. Are you okay with that being replaced back to simple urfav CLI flags?

nonsense:
> @nonsense - I think you added the config file structure. Are you okay with that being replaced back to simple urfav CLI flags?

I think we will end up with too many CLI flags. I don't think it is great UX when you have to specify a command line with 10 different options.

willscott:
new plan is to let urfavcli manage the file thing per https://github.com/urfave/cli/blob/master/docs/v2/manual.md

so flags will be source of truth, and it'll merge from file

willscott:
take a look, @nonsense

mvdan (approved):
I find this to be a lot saner and more consistent :)

Hard to say for sure that no typos or tiny mistakes were made, but we'd find them sooner than later anyway.

nonsense (approved):
lgtm

nonsense:
We have to update the README, but I don't mind doing that after the PR is merged and we try the new setup.

mvdan:
Merging since this will simplify running the controller+daemon for testing.

PR #38, opened 2021-04-14
integration tests: add storage deal task test with devnet

nonsense:
This PR is adding an integration test for the `dealbot` command for a storage task, assuming that `dealbot` and `lotus`-suite binaries have already been compiled.

It is still work-in-progress, because the first storage deal seems to always fail with:

```2021-04-14T15:24:57.297+0300    INFO    dealbot commands/storage_deal.go:85     Deal status     {"cid": "bafyreidft3zmdajfywxzyorqlmzkoczw2pn2i77mhuodmh6lab2bhl7a6y", "piece": "baga6ea4seaqmroj7ckz3ovo3mmgdejljoalak2ruhpp2gquux2ojzh7bctkoskq", "state": "StorageDealFailing", "message": "unexpected deal status while waiting for data request: 11 (StorageDealFailing). Provider message: deal rejected: proposed provider collateral above maximum: 799920 > 0", "provider": "f01000"}
2021-04-14T15:24:57.297+0300    INFO    dealbot commands/storage_deal.go:85     Deal stage      {"cid": "bafyreidft3zmdajfywxzyorqlmzkoczw2pn2i77mhuodmh6lab2bhl7a6y", "name": "StorageDealReserveClientFunds", "description": "Reserving client funds", "created": "2021-04-14T15:24:27.450186+03:00", "expected": "a few minutes", "updated": "2021-04-14T15:24:27.55696+03:00", "size": 2032, "duration": 519970, "deal_id": 0, "piece_cid": "baga6ea4seaqmroj7ckz3ovo3mmgdejljoalak2ruhpp2gquux2ojzh7bctkoskq", "message": "unexpected deal status while waiting for data request: 11 (StorageDealFailing). Provider message: deal rejected: proposed provider collateral above maximum: 799920 > 0", "provider": "f01000", "price": "500000000", "verfied": false}
2021-04-14T15:24:57.297+0300    INFO    dealbot commands/storage_deal.go:85     Deal stage      {"cid": "bafyreidft3zmdajfywxzyorqlmzkoczw2pn2i77mhuodmh6lab2bhl7a6y", "name": "StorageDealFundsReserved", "description": "FundsReserved", "created": "2021-04-14T12:24:57.297265Z", "expected": "a few minutes", "updated": "2021-04-14T12:24:57.297265Z", "size": 2032, "duration": 519970, "deal_id": 0, "piece_cid": "baga6ea4seaqmroj7ckz3ovo3mmgdejljoalak2ruhpp2gquux2ojzh7bctkoskq", "message": "unexpected deal status while waiting for data request: 11 (StorageDealFailing). Provider message: deal rejected: proposed provider collateral above maximum: 799920 > 0", "provider": "f01000", "price": "500000000", "verfied": false}
```

---

TODO:
- [x] confirm that the storage deal has completed successfully (either by checking logs, or by checking exit code)
- [x] extract the `data CID` from the storage deal and attempt to execute a `retrieve` task after that.

### DIFF
--- a/cmd/devnet/main.go
+++ b/cmd/devnet/main.go
@@ -1,0 +1,9 @@
+package main
+
+import (
+	"github.com/filecoin-project/dealbot/devnet"
+)
+
+func main() {
+	devnet.Main()
+}
--- a/devnet/devnet.go
+++ b/devnet/devnet.go
@@ -1,9 +1,7 @@
 package devnet
 
 import (
-	"bytes"
 	"context"
-	"fmt"
 	"log"
 	"os"
 	"os/exec"
@@ -14,46 +12,28 @@ import (
 )
 
 func runLotusNode(ctx context.Context) {
-	var stdout, stderr bytes.Buffer
 	lotusNodeCmd := "lotus-seed genesis new localnet.json  && lotus-seed pre-seal --sector-size 2048 --num-sectors 10 && lotus-seed genesis add-miner localnet.json ~/.genesis-sectors/pre-seal-t01000.json && lotus daemon --lotus-make-genesis=dev.gen --genesis-template=localnet.json --bootstrap=false"
 
 	cmd := exec.CommandContext(ctx, "sh", "-c", lotusNodeCmd)
-	cmd.Stdout = &stdout
-	cmd.Stderr = &stderr
-	err := cmd.Start()
+	_, err := cmd.CombinedOutput()
 	if err != nil {
-		log.Fatal(err)
-	}
-
-	err = cmd.Wait()
-	if err != nil {
-		log.Println(string(stderr.Bytes()))
-		log.Println("waiting on lotus node:", err)
+		log.Println(err.Error())
 	}
 }
 
 func runMiner(ctx context.Context) {
 	time.Sleep(5 * time.Second) // wait for lotus node to run
 
-	var stdout, stderr bytes.Buffer
 	lotusMinerCmd := "lotus wallet import ~/.genesis-sectors/pre-seal-t01000.key && lotus-miner init --genesis-miner --actor=t01000 --sector-size=2048 --pre-sealed-sectors=~/.genesis-sectors --pre-sealed-metadata=~/.genesis-sectors/pre-seal-t01000.json --nosync && lotus-miner run --nosync"
 
 	cmd := exec.CommandContext(ctx, "sh", "-c", lotusMinerCmd)
-	cmd.Stdout = &stdout
-	cmd.Stderr = &stderr
-	err := cmd.Start()
+	_, err := cmd.CombinedOutput()
 	if err != nil {
-		log.Fatal(err)
-	}
-
-	err = cmd.Wait()
-	if err != nil {
-		log.Println("waiting on lotus miner:", err)
+		log.Println(err.Error())
 	}
 }
 
 func publishDealsPeriodicallyCmd(ctx context.Context) {
-	var stdout, stderr bytes.Buffer
 	publishDealsCmd := "lotus-miner storage-deals pending-publish --publish-now"
 
 	for {
@@ -65,19 +45,15 @@ func publishDealsPeriodicallyCmd(ctx context.Context) {
 		}
 
 		cmd := exec.CommandContext(ctx, "sh", "-c", publishDealsCmd)
-		cmd.Stdout = &stdout
-		cmd.Stderr = &stderr
-		err := cmd.Start()
+		_, err := cmd.CombinedOutput()
 		if err != nil {
 			continue
 		}
 
-		cmd.Wait()
 	}
 }
 
 func setDefaultWalletCmd(ctx context.Context) {
-	var stdout, stderr bytes.Buffer
 	setDefaultWalletCmd := "lotus wallet list | grep t3 | awk '{print $1}' | xargs lotus wallet set-default"
 
 	for {
@@ -89,15 +65,7 @@ func setDefaultWalletCmd(ctx context.Context) {
 		}
 
 		cmd := exec.CommandContext(ctx, "sh", "-c", setDefaultWalletCmd)
-		cmd.Stdout = &stdout
-		cmd.Stderr = &stderr
-		err := cmd.Start()
-		if err != nil {
-			fmt.Println("error", err.Error())
-			continue
-		}
-
-		err = cmd.Wait()
+		_, err := cmd.CombinedOutput()
 		if err != nil {
 			continue
 		}
--- a/devnet/devnet.go
+++ b/devnet/devnet.go
@@ -1,0 +1,144 @@
+package devnet
+
+import (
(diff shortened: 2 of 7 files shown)


cmd/devnet/main.go
```diff
@@ -0,0 +1,9 @@
+package main
```
nonsense, line 1: We could certainly instantiate a `devnet` by using the lotus APIs and libraries directly in Go, but I think it is best if we test the dealbot with an actual version of the `lotus` binaries. WDYT?

nonsense:
@mvdan @willscott I know we want to avoid `bash`, but I think this test is certainly simpler than what I attempted to do with `testscript`.

I don't mind if we decide to refactor later and run as part of a `go test` suite. WDYT?

tasks/storage_deal.go
```diff
@@ -145,19 +139,23 @@ func MakeStorageDeal(ctx context.Context, config NodeConfig, node api.FullNode,
 			storagemarket.StorageDealSlashed,
 			storagemarket.StorageDealRejecting,
 			storagemarket.StorageDealFailing,
-			storagemarket.StorageDealError,
+			storagemarket.StorageDealError:
+
+			logStages(info, log)
+			return errors.New("storage deal failed")
 
 			// deal is on chain, exit successfully
```
mvdan, line 147: unindent?

integration_tests/header.sh
```diff
@@ -0,0 +1,27 @@
...
+set -x
+
+err_report() {
+    echo "Error on line $1 : $2"
+}
+#FILENAME=`basename $0`
+#trap 'err_report $LINENO $FILENAME' ERR
+
+function finish {
+  kill -9 $DEVNETPID
```
mvdan, line 13: shouldn't this be SIGINT instead of SIGKILL, to allow lotus etc to clean up after themselves? or do we not care, since we're in a temporary directory?

integration_tests/01_storage_retrieval_ok.sh
```diff
@@ -0,0 +1,38 @@
...
+
+TOKEN=`cat $LOTUS_PATH/token`
+API=`cat $LOTUS_PATH/api`
+DATADIR=$TEMPDIR
+MINER=t01000
+
+dealbot --api $TOKEN:$API storage-deal --data-dir $DATADIR --miner $MINER
+
+returnValue=$?
+if [ $returnValue -eq 0 ]; then
```
mvdan, line 17: since this is bash, might as well use `[[` since it's easier and safer :)

integration_tests/header.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
```
mvdan, line 1: do we want this script to do something like a `go install`, to ensure one is not running the tests with an old `dealbot` binary?

cmd/devnet/main.go
mvdan: Doing it all in pure Go would have the advantage that we don't need the user to do a separate build of lotus, and it would also give us the pinning of the lotus version for free. So I think it makes more sense in the long run. It can certainly be left as a TODO here, though.

devnet/devnet.go
```diff
@@ -0,0 +1,144 @@
...
+
+	cmd := exec.CommandContext(ctx, "sh", "-c", lotusNodeCmd)
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+	err := cmd.Start()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	err = cmd.Wait()
```
mvdan, line 28: nit: use CombinedOutput instead of Start/Wait with output buffers. Same result, and printing the output on error is still possible.

mvdan (commented):
should the non-test changes be split up into another PR?

overall LGTM, any test is better than no test. I do think we'll want to move this to `go test` soon :)

integration_tests/header.sh
nonsense: I don't think we care, since it is a temporary directory, we just want the process to release the socket.

integration_tests/header.sh
nonsense: I was thinking of doing that in `.circleci/config.yml` - not sure if the test itself should be responsible for the building of the dependencies.

nonsense:
> Doing it all in pure Go would have the advantage that we don't need the user to do a separate build of lotus, and it would also give us the pinning of the lotus version for free. So I think it makes more sense in the long run. It can certainly be left as a TODO here, though.

I am not sure I want the `dealbot` to be responsible for building `lotus`, `lotus-miner` and `lotus-seed` - especially when it is not a simple build process - it involves `filecoin-ffi`, etc.

I do agree that having the integration test itself in `go test` would be nice - that could be done on next iteration.

tasks/storage_deal.go
nonsense: Not sure what you mean, I usually leave that to `gofmt` / `goimports` - should I be using something else?

mvdan: it allows a comment in this line to either be indented with the body before, or with the case below. the comment logically refers to the comment below, so it should be unindented, like it. in some scenarios like this one, gofmt just follows your choice.

cmd/devnet/main.go
nonsense: I am not sure I want the dealbot to be responsible for building lotus, lotus-miner and lotus-seed - especially when it is not a simple build process - it involves filecoin-ffi, etc.

I do agree that having the integration test itself in go test would be nice - that could be done on next iteration.

integration_tests/header.sh
mvdan: personally I think CI should be as thin as possible; running the tests should be easy, akin to just `go test ./...`. Plus, a `go install` should be practically instantaneous if nothing changed :)

mvdan:
> I am not sure I want the `dealbot` to be responsible for building `lotus`, `lotus-miner` and `lotus-seed` - especially when it is not a simple build process - it involves `filecoin-ffi`, etc.

I guess we can stub what we can, like we already do in a go.mod replace for ffi?

If that isn't enough for a `go build` to just work for us, then yeah, it's going to be messy.

nonsense:
> I guess we can stub what we can, like we already do in a go.mod replace for ffi?

> If that isn't enough for a go build to just work for us, then yeah, it's going to be messy.

I think it is going to be messy, and I'd rather not do this as part of this PR. I welcome this effort, and I'd be very happy if it is successful, but we already made various attempts at this before, and it didn't end well.

integration_tests/header.sh
mvdan: Fair, we can leave that for a future "quality of life" improvement.

mvdan (approved):
LGTM. I assume you'll let us know when you've figured out that problem with the deals.

nonsense:
As a follow up I will see about adding this to be run in CI.

PR #40, opened 2021-04-14
update integration test to pass for both storage deals

nonsense:
Looks like the first storage deal actually passes on CI (but fails for me consistently when ran locally), so we can adjust the integration test for CI for now, until we figure out why it fails locally (possibly due to faster machine). WDYT?


integration_tests/01_storage_retrieval_ok.sh
```diff
@@ -14,8 +14,8 @@ MINER=t01000
 dealbot --api $TOKEN:$API storage-deal --data-dir $DATADIR --miner $MINER
```
nonsense, line 14: Alternatively we could just remove the 2nd storage deal - no need to wait for 2 storage deals on CI, it will just make the suite slower.

willscott (approved):

PR #44, opened 2021-04-15
Setup Metrics Interface

hannahhoward:
# Goals

Setup a framework for exporting metrics

# Implementation

- Define an abstract interface for observing tasks - rather than attempt to wrangle one of the Open* frameworks, I wrote a high level interface that can just record observations on a task when it updates.
- I wrote three implementations here:
   - One that will record metrics for Prometheus
   - One that will just generate logs for the controller
   - One that is just a test version used to assert observations
- The prometheus one needs further setup and testing -- I have yet to build a docker for prometheus itself, but this PR is growing quickly and I want to get feedback before I go further
- My assumption is: you'd run the log version in dev, the prometheus version in prod, allowing us to avoid having devs running prometheus either on their machine or on a test node
- I also updated the logic of the update task command so it accepts further tasks updates IF and only if the same deal bot that assigned itself submits the update
- I went ahead and wrote the a controller HTTP test that tests the interactions with the controller HTTP server, using the api client, absent the rest of the system. This seems like a good place to start in terms of isolated testing (though it sits between unit and integration testing, I think it's a good boundary to work with)

### DIFF
--- a/tasks/model.go
+++ b/tasks/model.go
@@ -43,9 +43,13 @@ const (
 	Failed
 )
 
-var StatusNames = map[Status]string{
+var statusNames = map[Status]string{
 	Available:  "Available",
 	InProgress: "InProgress",
 	Successful: "Successful",
 	Failed:     "Failed",
+}
+
+func (s Status) String() string {
+	return statusNames[s]
 }
--- a/tasks/model.go
+++ b/tasks/model.go
@@ -1,6 +1,8 @@
 package tasks
 
 import (
+	"time"
+
 	"github.com/filecoin-project/go-address"
 	logging "github.com/ipfs/go-log/v2"
 )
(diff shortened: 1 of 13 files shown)


config/env.go
```diff
@@ -7,7 +7,8 @@ type EnvConfig struct {
 }
 
 type ControllerConfig struct {
-	Listen string `toml:"listen"`
+	Listen  string `toml:"listen"`
+	Metrics string `toml:"metrics"` // what to use to record metrics -- etiher "log", or "prometheus"
```
willscott, line 11: pls look at #37

willscott (commented):
the observer setup looks very reasonable to me on first readthrough.

I didn't see the set of labels as too verbose, and would probably lean on the verbose rather than terse side initially.

tasks/model.go
```diff
@@ -40,3 +42,10 @@ const (
 	Successful
 	Failed
 )
+
+var StatusNames = map[Status]string{
```
mvdan, line 46: use stringer? :)

willscott (approved):

tasks/model.go
hannahhoward: makes sense

PR #45, opened 2021-04-15
fix nil ptr dereference

nonsense:
This PR is:
1. Fixing a nil pointer dereference.
2. Fixing the description of the `start-offset` flag.
3. Adding two log lines so that we debug which Lotus API is blocking for an unknown reason.

### DIFF
--- a/tasks/storage_deal.go
+++ b/tasks/storage_deal.go
@@ -109,11 +109,15 @@ func MakeStorageDeal(ctx context.Context, config NodeConfig, node api.FullNode, 
 		return err
 	}
 
+	log("got proposal cid", "cid", proposalCid)
+
 	// track updates to deal
 	updates, err := node.ClientGetDealUpdates(ctx)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


tasks/storage_deal.go
```diff
@@ -156,6 +160,10 @@ func MakeStorageDeal(ctx context.Context, config NodeConfig, node api.FullNode,
 }
 
 func logStages(info api.DealInfo, log UpdateStatus) {
+	if info.DealStages == nil {
+		log("Deal stages is nil")
```
mvdan, line 164: This would still panic, no? Needs a return of some sort.

mvdan (approved):

tasks/storage_deal.go
nonsense: 🤦 of course.

We could probably remove this after a while, but there are still a lot of clients/miners running older versions where `DealStages` is nil.

PR #46, opened 2021-04-15
upgrade lotus to current master

nonsense:
I decided to upgrade Lotus before attempting to understand what is wrong with `node.ClientGetDealUpdates` and why it blocks forever without returning a channel for updates for a given deal, when the dealbot targets mainnet.


mvdan (approved):
LGTM, but I wonder if it would be better to just stick to the latest stable tag unless we have a good reason to go beyond that.

nonsense:
> unless we have a good reason to go beyond that. 

`node.ClientGetDealUpdates` seems to block forever when trying a deal towards `mainnet` - this is why I am upgrading Lotus, to make sure if it is fixed/changed, and if not - report the bug or try to fix it.

PR #47, opened 2021-04-15
allow running integration tests against an existing lotus node

mvdan:
(see commit message)


willscott (approved):

integration_tests/header.sh
```diff
@@ -1,33 +1,35 @@
 #!/bin/bash
 
+my_dir="$(dirname "$0")"
+
 #set -o errexit ... re-enable when we fix the storage deal bug on Lotus side
 set -x
 
-err_report() {
```
nonsense, line 6: Now that tests are passing, we might want to revert this and also set `set -e` so that we exit when a command returns exit code != 0

mvdan: Sure :) I just assumed that this function was unused. I guess one also needs a trap line?

PR #50, opened 2021-04-16
fix(devnet): fix tunnel.sh shell script on mac

hannahhoward:
# Goals

I hate shell scripting in compatibilities but...

for some reason on mac the capture ssh command puts array index 0 as blank, 1 as the API, and 2 as
the token

# Implementation
For the time being I solved by running awk twice.

I'm sure there's a better solution

My bash foo is weak.


willscott (approved):

mvdan:
Hm, unfortunate. I used awk precisely to avoid having to ssh twice. If you do that, might as well just use `cat` :)

willscott (approved):

mvdan (approved):

PR #52, opened 2021-04-16
add sshfs to tunnel script for storage deals

mvdan:
(see commit message)

### DIFF
--- a/devnet/tunnel.sh
+++ b/devnet/tunnel.sh
@@ -5,7 +5,16 @@ ADDRESS=ubuntu@ec2-3-83-188-192.compute-1.amazonaws.com
 # Start the tunnel to port forward the lotus API on :1234.
 # This is a no-op if the port is already running.
 echo "Connecting tunnel..."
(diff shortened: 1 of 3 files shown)


willscott (approved):
maybe spit out a warning / error if sshfs isn't installed?

or at least `set -e` to fail at that point

gammazero (approved):
LGTM
It would be nice to have a convenient way to shutdown sshfs and the tunnel, but not necessary for now.

mvdan:
> maybe spit out a warning / error if sshfs isn't installed?

The script should now exit if either of the first two commands fails. The third one is inside a command substitution, so that's trickier.

> It would be nice to have a convenient way to shutdown sshfs and the tunnel, but not necessary for now.

You're right; for now I've found that `killall ssh` is enough for my needs. I'm not sure if there could be a cleaner way. I guess we could teach the script to stay running in the foreground, and stop all the sub-processes on ^C, but then you lose all the `source` goodness with env vars.

PR #55, opened 2021-04-16
Add a keypair identity that signs records to the controller

willscott:


tasks/model.go
```diff
@@ -24,6 +25,19 @@ type Task struct {
 	StorageTask   *StorageTask   `json:"storage_task,omitempty"`
 }
 
+func (t Task) Bytes() []byte {
+	b, err := json.Marshal(&t)
+	if err != nil {
+		return []byte{}
```
mvdan, line 31: This worries me a bit - if for some reason we get a json error, we'd then happily send an empty task, signing those empty bytes too.

We just have one call site, so I'd do the json.Marshal and double error check there. If we find that too tedious and repeated in a bunch of places, we could add a function or method on an entire AuthenticatedTask, like Sign or Verify.

mvdan: Here's another detail to consider - json.Marshal isn't canonical/stable, and it's certainly possible that the encoding would change over time e.g. if encoding/json gets bugfixes or if we change the struct type.

Can this possibly affect us? E.g. if we signed the task when it was assigned, upgrade the controller while it's in progress, and later end up with different json-encoded bytes when we want to verify a signature.

If you think that might be an issue, one possible solution would be to store the originally signed bytes next to the signature. Then we're also free to change the encoding entirely, for example.

controller/controller.go
```diff
@@ -39,11 +43,45 @@ func New(ctx *cli.Context) (*Controller, error) {
...
+		if err != nil {
+			return nil, err
+		}
+
+		// save it.
+		b, err := crypto.MarshalPrivateKey(pr)
+		if err != nil {
+			return nil, err
+		}
+		if err := ioutil.WriteFile(identity, b, 0666); err != nil {
```
mvdan, line 63: this is a private key, so consider 0600 perhaps

controller/state.go
```diff
@@ -11,87 +11,148 @@ import (
...
 				if t.WorkedBy != req.WorkedBy {
-					return errors.New("task already acquired")
+					return nil, errors.New("task already acquired")
 				}
 			}
 			log.Infow("state update", "uuid", t.UUID, "status", req.Status, "worked_by", req.WorkedBy)
 
 			t.Status = req.Status
+			var err error
+			t.Signature, err = key.Sign(t.Bytes())
```
mvdan, line 59: There's no commit message here, so I'm assuming that what you want to do is sign each task with a private key, handing the signature to the daemon grabbing the task, so that only that same daemon can later report updates on the task.

If another client tries to update that task, they would fail as it would be practically impossible for them to guess the same signature.

If my understanding is correct, this feels wrong - we should only create a signature when the task is assigned (next to StartedAt etc), and in the "else" we should verify the signature.

mvdan (commented):
I tried to only review your changes here.

tasks/model.go
willscott: eventually we probably want this over the data in ipld schema (once we get to reputation system). I suspect this code changes a couple more times as we move in that direction

controller/state.go
willscott: nope. this is for external downstream consumers of the dealbot system (bots+controller) to be able to validate that the reports did come from this dealbot system

controller/controller.go
willscott: fixed in https://github.com/filecoin-project/dealbot/pull/57/commits/1e73b9e2ad2a0bb05c69ce267450609727cd24ef

controller/state.go
mvdan: Oh, I see. How do we ensure that only the right client can update a task, then - by assuming that they would otherwise not be able to guess the UUID? I seem to recall that our original UUIDs were really short, though.

mvdan:
SGTM. Closing as we've folded into #57.

PR #56, opened 2021-04-16
Add task ingestion

hannahhoward:
# Goals

Add the ability to add storage and retrieval tasks to the queue for the controller. This is a simple implementation that allows adding a single task, either storage or retrieval. I can make batch endpoints to add multiple storage and retrieval tasks at the same time as well if it is useful.

# Implementation

- I did a bit of rejiggering of the API to make it more closely follow REST rules (apologies, old web programmer here who gets picky about GET/PUT being idempotent :)
- I added a simple add a strorage task and add a retrieval task endpoint. The other task parameters are controlled by the controller so it made sense to put them in two seperate endpoints that just take the specific parameters. If it's important i can make it a combined endpoint
- Add a get task buy UUID endpoint -- seems like this will be important at some point
- Rejigger HTTP tests to add more functionality testing.

Note: This does not actually persist anything -- it's just adding to the currently in memory queue.

### DIFF
--- a/controller/client/client.go
+++ b/controller/client/client.go
@@ -16,6 +16,14 @@ import (
 
 var log = logging.Logger("controller")
 
+type ErrRequestFailed struct {
+	Code int
+}
+
+func (e ErrRequestFailed) Error() string {
+	return fmt.Sprintf("Request failed: %s", http.StatusText(e.Code))
+}
+
 // Client is the API client that performs all operations
 // against the dealbot controller.
 type Client struct {
@@ -49,13 +57,13 @@ func (c *Client) Close() error {
 }
 
 func (c *Client) ListTasks(ctx context.Context) ([]*tasks.Task, error) {
-	reader, _, err := c.request(ctx, "GET", "/tasks", nil)
+	resp, err := c.request(ctx, "GET", "/tasks", nil)
 	if err != nil {
 		return nil, err
 	}
 
 	var res []*tasks.Task
-	err = json.NewDecoder(reader).Decode(&res)
+	err = json.NewDecoder(resp.Body).Decode(&res)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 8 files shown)


controller/tasks.go
```diff
@@ -34,12 +38,94 @@ func (c *Controller) updateTaskHandler(w http.ResponseWriter, r *http.Request) {
...
+
+	task, err := State.NewStorageTask(storageTask)
+	if err != nil {
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
+
+	taskURL, err := r.URL.Parse("/tasks/" + task.UUID)
+	if err != nil {
+		w.WriteHeader(http.StatusBadRequest)
```
willscott, line 74: return?

controller/tasks.go
```diff
@@ -34,12 +38,94 @@ func (c *Controller) updateTaskHandler(w http.ResponseWriter, r *http.Request) {
...
+	err := json.NewDecoder(r.Body).Decode(&storageTask)
+	if err != nil {
+		log.Errorw("StorageTask json decode", "err", err.Error())
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
+
+	task, err := State.NewStorageTask(storageTask)
+	if err != nil {
+		w.WriteHeader(http.StatusBadRequest)
```
willscott, line 68: log the error in these cases?

controller/client/client.go
```diff
@@ -49,33 +57,116 @@ func (c *Client) Close() error {
...
+}
+
+func (c *Client) CreateRetrievalTask(ctx context.Context, retrievalTask *tasks.RetrievalTask) (*tasks.Task, error) {
+	var body bytes.Buffer
+	err := json.NewEncoder(&body).Encode(retrievalTask)
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := c.request(ctx, "POST", "/tasks/retrieval", bytes.NewReader(body.Bytes()))
```
mvdan, line 149: I realised I posted this in @willscott's PR, because I had not noticed the PRs were atop each other.

You can use the `body` buffer directly in these :)

mvdan (approved):
SGTM. We should get this merged sooner than later to not block other work.

PR #57, opened 2021-04-18
DB for state

### LINKED ISSUE #5: Persistance for active deals
The process should have a way to track the deals it has in flight, so that if it is restarted / re-deployed it can continue tracking them.

There are a couple options on the top of my head to manage this persistence:
* we could have the process run with access to a persistent volume / storage location where this in-flight information is stored.
* we could structure this 'state' object to remained synced with a driving 'orchestrator' program we expect to eventually exist
  * for instance: the orchestrator in asking for a deal can assign a uuid to the request. once a deal id or state object exists for it, the dealbot can response to save that state object with the uuid. If the orchestrator later asks the bot about a request uuid and the bot has been redeployed ad doesn't know about it, it can request the state back from the orchestrator and resume where it left off.  Other designs may turn out preferable / easier

willscott:
This PR adds a sql abstraction behind the controller that should support either sqlite or postgres.

The final commit in this branch hold all the sql work. The previous ones are merging of #55 and #56

fix #5


mvdan (approved):
Overall SGTM.

@hannahhoward 's PR had a few comments in https://github.com/filecoin-project/dealbot/pull/56, do we want to deal with those post-merge to allow master to be useful again?

I also haven't thoroughly reviewed @gammazero's changes, but they look fine - and anything else we can fix in follow-ups, IMO.

I'll merge after lunch if noone beats me to it or objects.

mvdan:
A rebase-merge causes conflicts, so I'm going for a merge commit, for the sake of keeping history.

PR #60, opened 2021-04-21
controller: introduce a "pop task" endpoint

mvdan:
(see commit message)

### DIFF
--- a/controller/tasks.go
+++ b/controller/tasks.go
@@ -27,6 +27,42 @@ func (c *Controller) getTasksHandler(w http.ResponseWriter, r *http.Request) {
 	json.NewEncoder(w).Encode(tasks)
 }
 
+func (c *Controller) popTaskHandler(w http.ResponseWriter, r *http.Request) {
+	logger := log.With("req_id", r.Header.Get("X-Request-ID"))
+
+	logger.Debugw("handle request", "command", "pop task")
+	defer logger.Debugw("request handled", "command", "pop task")
+
+	w.Header().Set("Content-Type", "application/json")
+
+	// TODO: use a single SQL transaction to query the oldest available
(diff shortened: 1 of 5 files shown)


mvdan:
@gammazero the TODO is pretty relevant to your work :) It's what I briefly mentioned on the standup yesterday; being able to grab an available task via a single query.

controller/tasks.go
```diff
@@ -27,6 +27,42 @@ func (c *Controller) getTasksHandler(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+	var firstAvailable *tasks.AuthenticatedTask
+	for _, task := range allTasks {
+		if task.Status == tasks.Available {
+			firstAvailable = task
+			break
+		}
+
+	}
```
willscott, line 61: are we worried about startup dynamics where all clients spin up, ask for a task (get the same one), fail to claim it because a different one did already, and repeat? we can probably not worry about this via a small random time pause before asking for a task on startup, or could rotate the list so that recently asked for tasks get moved to the back of the range.

mvdan: You're right that this mechanism is very simple and will likely lead to races. This is why the TODO is there :) I think the right solution is to use a select+update transaction that grabs and assigns the oldest available task at the DB level. That can be done on top of @gammazero's work to split into two tables, once that lands.

gammazero: Do you want to keep this PR open and either:
1) Have me merge my PR into this and then complete the TODO here?
2) Keep this PR open and rebase it once my upcoming PR is in?

mvdan: I don't think the work involved in either of those is necessary :) this PR doesn't break or change anything in a bad way, so we can merge into master and then have your change merge the new master.

PR #63, opened 2021-04-21
Enable local testing with fake dealbot

hannahhoward:
# Goals

Provide a local setup that does not require Lotus at all

# Implementation

- Enable a full local testing node by using a deal bot that simply reports statuses without making
deals. 
- With this mock dealbot, we can setup a docker-compose stack that will run everything up to
prometheus (along with postgres) so that you can work with all parts of the stack that are not
lotus. 
- Also adds command lines for queue-ing storage and retrieval in the controller at the command line.
- The mock dealbot can be adjusted to produce different rates of success or failure and also shorter or longer task times.

# For Discussion

The one downside here is the work to maintain the fake dealbot along with the real one... perhaps there is a way to make it lest duplicative? Will think about it.


docker-compose.yml
```diff
@@ -0,0 +1,36 @@
...
+  controller:
+    build: .
+    command: ["controller", "--listen", ":8764", "--metrics", "prometheus", "--driver", "postgres", "--dbloc", "postgresql://admin:dealbot-admin@db:5432/dealbot?sslmode=disable"]
+    ports:
+      - "8764:8764"
+    depends_on: 
+      - db
+  dealbot:
+    build: .
+    command: ["mock", "--endpoint", "http://controller:8764"]
```
willscott, line 19: since this is just a testing compose setup, we should probably give it a more appropriate name?

hannahhoward: sure!

willscott (approved):

Issue #64, opened 2021-04-22
Remove default  tasks from controller


willscott:
when testing, it'll be useful if we can fully control the ingest to the controller

mvdan:
I was about to suggest this too :) I want the controller-daemon integration test to be fairly self-isolated.

PR #65, opened 2021-04-22
extract initial sample tasks to a script

willscott:
move the initial 4 sample tasks out to a separate bash script. no tests appear to make use of them yet, and having them as a separate command will make testing more predictable.

cc @gammazero


hannahhoward (approved):

willscott:
fix #64

PR #67, opened 2021-04-23
Make pop-task non-racy

hannahhoward:
# Goals

While we wait for an SQL rearchitect, this PR implements the machinery to make pop-task non-racy and uses a simple mutex for now to protect the database.

# Implementation

- Convert pop-task to POST
- Have it take a WorkedBy in the request body
- Use a mutex to protect Pop-task handler (remove once we have an SQL version)
- Update the task status to InProgress and assign it in the pop task handler

### DIFF
--- a/controller/client/client.go
+++ b/controller/client/client.go
@@ -76,8 +76,14 @@ func (c *Client) ListTasks(ctx context.Context) ([]*tasks.Task, error) {
 	return res, nil
 }
 
-func (c *Client) PopTask(ctx context.Context) (*tasks.Task, error) {
-	resp, err := c.request(ctx, "GET", "/pop-task", nil)
+func (c *Client) PopTask(ctx context.Context, r *PopTaskRequest) (*tasks.Task, error) {
+	var body bytes.Buffer
+	err := json.NewEncoder(&body).Encode(r)
+	if err != nil {
+		return nil, err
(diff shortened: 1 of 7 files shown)


willscott (approved):

gammazero (commented):
I do not believe this will not protect against races in the DB, at least not with postgres. You will need a transaction for this.  My upcoming changes do transactionalize this query/update, so unless this cures an immediate problem with sqlite, then maybe better to wait.

mvdan:
I agree that this won't get rid of all DB races - one could still have a race between `POST /pop-task` and a `PATCH /tasks/{same-uuid}`, for example. But I also don't think it hurts and I agree with Hannah that it makes testing easier to not have multiple `pop-task` calls race with each other. The global lock is making them be serial, at least.

@gammazero by all means remove the lock and the separate SELECT/UPDATE when your transaction stuff is ready - but for now I don't see a problem with merging while that isn't ready, and it looks like the conflicts will be rather minimal. And the change towards POST is something we want to do anyway :)

mvdan (approved):

hannahhoward:
@gammazero this is still intermediate solution -- yes there are other races and the DB solution is the endpoint but we need your changes first and this makes things much easier.

PR #69, opened 2021-04-23
integration_tests: add first test including the controller

mvdan:
(see commit message)

### DIFF
--- a/integration_tests/header.sh
+++ b/integration_tests/header.sh
@@ -3,7 +3,6 @@
 my_dir="$(dirname "$0")"
 
 #set -o errexit ... re-enable when we fix the storage deal bug on Lotus side
-set -x
 
 # go install dealbot and devnet
 pushd $my_dir/../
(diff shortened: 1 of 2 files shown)


mvdan:
Also, it would help if one of you can run it locally and confirm that it works for you too.

We should also add it to CI; I didn't do that yet because each script starts its own devnet, and that's really slow. We should first refactor `header.sh` to start one devnet for all the integration tests.

integration_tests/header.sh
```diff
@@ -3,7 +3,6 @@
 my_dir="$(dirname "$0")"
 
 #set -o errexit ... re-enable when we fix the storage deal bug on Lotus side
-set -x
```
willscott, line 6: don't we want this?

willscott (approved):

integration_tests/header.sh
mvdan: See the last part of the commit message.

mvdan:
This is old and I have more changes now, such as a better devnet and the ability to tunnel into the new AWS machine. Closing and will open a new PR.

PR #70, opened 2021-04-24
New DB schema to split tasks from status history

gammazero:
- DB transactions
- Task history
- Pop-task atomically finds available task and assigns it


willscott (approved):
👍

Issue #62, opened 2021-04-21
Dashboards Data Collection Prioritization


mgoelzer:
Let the set ActiveMiners = `{ all miners on chain minus those that have not made a deal in the past 60 days or have zero wallet balance] }`

### MVP Dashboards Data Collection:  Golden Path (32 GiB deals from a verified/datacap wallet)

(assuming only one geo of dealbots at this stage)

 - For each element in ActiveMiners:
   - Is the miner reachable at the libp2p level?  (Does it have a peer id?)  If not, log a failure of type "miner offline" and `continue` this loop
   - Is the miner even reachable for things like `query-ask`.  If not, log a failure of type "miner not accepting deals" and `continue` loop
   - Is the miner charging greater than `0 FIL/GB` for **verified/datacap** storage?  If yes, log a WARN of type "verified storage price > 0" and keep going on with this iteration of the loop (no `continue`)
   - Is the miner charging greater than `0.01 FIL/GB` for **verified/datacap** storage?  If yes, log a failure of type "verified storage price too expensive (>0.01 FIL/GB)" and keep going
   - Ditto for verified retrieval prices more than `0.01 FIL/GB`
   - Attempt to make a 32 GiB (2^35 bytes exactly) storage with this miner; make sure  **verified/datacap** wallet, with **`--verified-deal` set**, and **`--fast-retrieval` set** [`*`]
     - On failure, retry up to 2 more times.  With a delay in between - maybe 6-12 hours?  If all fail, `continue` this loop
   - Wait 24 hours and attempt to retrieve the 32 GiB [`*`] 
     - Again, retry failures up to 2 more times with a delay.
 
 [`*`] We want to log both whether the _transfer_ was successful and whether the deal reached the final status of being active on chain (`StorageDealActive` I think is the client-side state for this).  For dashboards, we want DSRs for golden path deals (verified + 32 GiB) for both successful transfer and reaching the final `StorageDealActive` state.  On retrievals, we want to log both (1) the time between deal initiation and receipt of the first byte of data, and (2) the overall transfer speed in GiB per second.

### v0.2 Dashboards Data Collection:  Permutations of Different Deal Parameters

(still assuming only one geo)

Everything in MVP + add these additional test conditions against every miner:
 - Try verified and unverified deals, both 32 GiB
 - Try `--fast-retrieval` and non-fast retrieval deals, both with 32 GiB
[...]

PR #73, opened 2021-04-30
Much more detailed state updating for deals!

hannahhoward:
# Goals

Track all the things for deals

# Implementation

- Add the concept of stages (borrowed from go-fil-markets) to deals
- Update the state db to have a stage name on task_status_ledger
- Add updating of the task ledger in update task
- Implement much closer stage tracking in deal stages

### DIFF
--- a/commands/retrieval_deal.go
+++ b/commands/retrieval_deal.go
@@ -29,7 +29,7 @@ func makeRetrievalDeal(cctx *cli.Context) error {
 
 	log.Infof("remote version: %s", v.Version)
 
-	carExport := true
+	carExport := false
 	payloadCid := cctx.String("cid")
 
 	log.Infof("retrieving cid: %s", payloadCid)
@@ -43,18 +43,16 @@ func makeRetrievalDeal(cctx *cli.Context) error {
 		CARExport:  carExport,
 	}
 
-	err = tasks.MakeRetrievalDeal(cctx.Context, nodeConfig, node, task,
-		func(string, *tasks.StageData) error {
-			return nil
-		},
-		func(msg string, keysAndValues ...interface{}) {
-			log.Infow(msg, keysAndValues...)
-		})
+	err = tasks.MakeRetrievalDeal(cctx.Context, nodeConfig, node, task, emptyUpdateStage, log.Infow)
 	if err != nil {
 		log.Fatal(err)
 	}
 
 	log.Info("successfully retrieved")
 
+	return nil
+}
+
+func emptyUpdateStage(string, *tasks.StageDetails) error {
 	return nil
 }
--- a/commands/retrieval_deal.go
+++ b/commands/retrieval_deal.go
@@ -43,9 +43,13 @@ func makeRetrievalDeal(cctx *cli.Context) error {
 		CARExport:  carExport,
 	}
 
-	err = tasks.MakeRetrievalDeal(cctx.Context, nodeConfig, node, task, func(msg string, keysAndValues ...interface{}) {
-		log.Infow(msg, keysAndValues...)
-	})
+	err = tasks.MakeRetrievalDeal(cctx.Context, nodeConfig, node, task,
+		func(string, *tasks.StageData) error {
+			return nil
+		},
+		func(msg string, keysAndValues ...interface{}) {
+			log.Infow(msg, keysAndValues...)
+		})
 	if err != nil {
 		log.Fatal(err)
 	}
--- a/controller/client/model.go
+++ b/controller/client/model.go
@@ -2,14 +2,23 @@ package client
 
 import "github.com/filecoin-project/dealbot/tasks"
 
+// PopTaskRequest specifies parameters to pop the next assigned task
 type PopTaskRequest struct {
-	Status   tasks.Status `json:"status"`
-	WorkedBy string       `json:"worked_by,omitempty"` // which dealbot works on that task
+	// Status indicates the status the task should be in after assignment
+	Status tasks.Status `json:"status"`
+	// WorkedBy is the dealbot that will be assigned to the task
+	WorkedBy string `json:"worked_by,omitempty"`
 }
 
+// UpdateTaskRequest specifies parameters for updating a task
 type UpdateTaskRequest struct {
-	Status    tasks.Status `json:"status"`
-	StageName string
-	StageData *tasks.StageData
-	WorkedBy  string `json:"worked_by,omitempty"` // which dealbot works on that task
+	// Status is the global task status, shared among all tasks types -- always one of Available, InProgress, Successful, or Failure
+	Status tasks.Status `json:"status"`
+	// Stage is a more detailed identifier for what part of the process the deal is in. Some stages are unique to the task type
+	// When a task status is Successful or Failure, Stage indicates the final stage of the deal making process that was reached
+	Stage string
+	// CurrentStageDeatails offers more information about what is happening in the current stage of the deal making process
+	CurrentStageDetails *tasks.StageDetails
+	// WorkedBy indicates the dealbot making this update request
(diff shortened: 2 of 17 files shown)


controller/client/model.go
```diff
@@ -2,7 +2,14 @@ package client
...
 
-type UpdateTaskRequest struct {
+type PopTaskRequest struct {
 	Status   tasks.Status `json:"status"`
 	WorkedBy string       `json:"worked_by,omitempty"` // which dealbot works on that task
 }
+
+type UpdateTaskRequest struct {
+	Status    tasks.Status `json:"status"`
+	StageName string
```
gammazero, line 12: I think a comment on what the fields are used for would be great.

I am guessing that `stageName` identifies a stage in task processing, and is not an actor's alias.  However, it is not clear how/if it relates to `Status`.

tasks/model.go
```diff
@@ -77,3 +94,80 @@ var statusNames = map[Status]string{
...
+	},
+	"FirstByteReceived": {
+		Description:      "First byte of data received from miner",
+		ExpectedDuration: "a few seconds, or several hours when unsealing",
+	},
+	"DealComplete": {
+		Description:      "All bytes received and deal is completed",
+		ExpectedDuration: "a few seconds",
+	},
+}
```
gammazero, line 136: When a deal is scheduled to run in the future but is not yet running, do we want a stage to indicate this?  Maybe having a NULL `StageName` for an InProgress task is enough?

Is there any way a deal can be cancelled before it is complete?

commands/retrieval_deal.go
```diff
@@ -43,9 +43,13 @@ func makeRetrievalDeal(cctx *cli.Context) error {
 		CARExport:  carExport,
 	}
 
-	err = tasks.MakeRetrievalDeal(cctx.Context, nodeConfig, node, task, func(msg string, keysAndValues ...interface{}) {
-		log.Infow(msg, keysAndValues...)
-	})
+	err = tasks.MakeRetrievalDeal(cctx.Context, nodeConfig, node, task,
+		func(string, *tasks.StageData) error {
```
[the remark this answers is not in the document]
hannahhoward: this one does nothing cause this is the isolated deal execution command -- it does not send updates back to the controller.

tasks/retrieval_deal.go
```diff
@@ -17,18 +18,49 @@ type RetrievalTask struct {
...
+			log:    log,
+		},
+		task: task,
+	}
+
+	err := executeStage("MinerOnline", updateStage, []step{
+		{de.getTipSet, "Tipset successfully fetched"},
+		{de.getMinerInfo, "Miner Info successfully fetched"},
+		{de.getPeerAddr, "Miner address validated"},
+		{de.netConnect, "Connected to miner"},
```
gammazero, line 37: What is the reason for the hierarchy of steps within a stage, instead of having each step as a separate stage?  Would it might be simpler for a retrieval deal to have the following stages:
- GetTipSet
- GetMinerInfo
- GetPeerAddr
- NewConnect
- QueryOffer

tasks/retrieval_deal.go
```diff
@@ -37,30 +69,98 @@ func MakeRetrievalDeal(ctx context.Context, config NodeConfig, node api.FullNode
...
+
+			AddLog(&dealStage, "file read from file system")
+			err = updateStage("DealComplete", &dealStage)
+			if err != nil {
+				return err
+			}
+
+			//if carExport {
+			//rdata = ExtractCarData(ctx, rdata, rpath)
+			//}
```
gammazero, line 160: Should probably remove this, or make it return `ErrNotImplemented`.

tasks/storage_deal.go
```diff
@@ -145,21 +266,37 @@ func MakeStorageDeal(ctx context.Context, config NodeConfig, node api.FullNode,
...
 	return nil
 }
 
-func logStages(info api.DealInfo, log UpdateStatus) {
+func toStageData(stage *storagemarket.DealStage) *StageData {
+	logs := make([]*Log, 0, len(stage.Logs))
+	for _, log := range stage.Logs {
+		logs = append(logs, &Log{
+			Log:       log.Log,
+			UpdatedAt: log.UpdatedTime.Time(),
```
gammazero, line 288: Better to copy the entire `Log` instead of specific fields, since this could be overlooked when adding a field to `Log`:
```go
logs := make([]*Log, len(stage.Logs))
for i, log := range stage.Logs {
    logCopy := *log
    logs[i] = &logCopy
}
```

gammazero (changes requested):

controller/http_test.go
```diff
@@ -101,14 +100,14 @@ func TestControllerHTTPInterface(t *testing.T) {
 			// when no tasks are available, pop-task should return nil
 			allTasks, err := apiClient.ListTasks(ctx)
 			require.NoError(t, err)
-			for _ = range allTasks {
-				_, err := apiClient.PopTask(ctx, &client.UpdateTaskRequest{
+			for range allTasks {
+				_, err := apiClient.PopTask(ctx, &client.PopTaskRequest{
```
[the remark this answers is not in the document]
hannahhoward: this is old code actually -- I just fixed it to work with what we're doing. But I'll go ahead and fix as needed

tasks/retrieval_deal.go
hannahhoward: the stages here were set by Mike's desires for how we mark things in the dashboards

tasks/retrieval_deal.go
hannahhoward: yea agree it's a copy paste but will change

tasks/storage_deal.go
hannahhoward: you can't do that because it's a different data structure

tasks/retrieval_deal.go
hannahhoward: so essentially, the fine grained-ness i'm trying to get to is exactly what's outlined here: https://github.com/filecoin-project/dealbot/issues/62

tasks/model.go
hannahhoward: yea null = in progress but not running. Also this will be a subsecond amount of time.

tasks/storage_deal.go
gammazero: Ah, yeah.  Oops.

controller/state/statedb.go
```diff
@@ -22,6 +21,15 @@ import (
...
 var log = logging.Logger("controller-state")
 
+type errorString string
+
+func (e errorString) Error() string {
+	return string(e)
+}
+
+const ErrNotAssigned = errorString("tasks must be acquired through pop task")
+const ErrWrongWorker = errorString("task already acquired by other worker")
```
gammazero, line 31: This is fine, but I was curious why do this above, and not
```go
var (
    ErrNotAssigned = errors.New("tasks must be acquired through pop task")
    ErrWrongWorker = errors.New("task already acquired by other worker")
)
```
as this allows you to compare `if err == ErrNotAssigned { ...`

gammazero (approved):
LGTM - with one comment/question

Also note test failure:
```
2021-04-30T21:49:06.015Z	FATAL	dealbot	commands/retrieval_deal.go:48	car export not implemented
```

controller/state/statedb.go
hannahhoward: that will still work. (arguably it's a slightly stricter form of equality but that also is good I think)

the main difference here it's a way to actually make errors constant. It drives me nuts that any program anywhere can reassign the values of sentinel errors in go.

hannahhoward: see: https://dave.cheney.net/tag/errors

PR #74, opened 2021-04-30
Add DB migrations

gammazero:
- Support DB migrations for both postgres and sqlite
- All DDL should go into migrations files, which are run in order.


willscott (approved):

go.mod
```diff
@@ -26,3 +27,5 @@ require (
 )
 
 replace github.com/filecoin-project/filecoin-ffi => github.com/filecoin-project/ffi-stub v0.1.0
+
+replace github.com/golang-migrate/migrate/v4 => github.com/gammazero/migrate/v4 v4.14.2-0.20210430025954-4d2c08f09e56
```
mvdan, line 31: why do we need this? document it?

gammazero: Done

PR #75, opened 2021-05-01
Setup CORS For basic auth

hannahhoward:
# Goals

Allow basic autho on controller endpoints from dashboard observable

# Implementation

- Setup cors responses to OPTIONS request
- Found some issues with the SQL setup code
- Add reverse proxy to docker to simulate basic auth - not there is some unusual configuration we'll need to support on the server
- Verified I was able to make a basic auth request from Observable against the docker setup

### DIFF
--- a/integration_tests/htpasswd/localhost-no-options
+++ b/integration_tests/htpasswd/localhost-no-options
@@ -1,0 +1,1 @@
+dealbot:$apr1$1sg1dh5n$bPXhILhAAbwSjsk1VdA5q/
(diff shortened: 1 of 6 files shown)


willscott (approved):

integration_tests/htpasswd/localhost-no-options
```diff
@@ -0,0 +1 @@
+dealbot:$apr1$1sg1dh5n$bPXhILhAAbwSjsk1VdA5q/
```
willscott, line 1: what is this password?

PR #77, opened 2021-05-02
attempt schema generation for dealbot data

willscott:


tasks/model.go
```diff
@@ -24,90 +23,39 @@ type NodeConfig struct {
...
+	InProgress Status = &_Status{x: 2}
 	// Successful means the task completed successfully
-	Successful
+	Successful Status = &_Status{x: 3}
 	// Failed means the task has failed
-	Failed
+	Failed Status = &_Status{x: 4}
 )
 
+func (f *_Status__Prototype) Of(x int) Status {
```
mvdan, line 44: I'd make this a standalone func, like `StatusOf`, because this assumes a little too much about the codegen. You also don't need it to be a method, I think.

An advantage of not relying on codegen details is that we could later move to node reflection without having to rewrite this func, for example :)

tasks/model.go
```diff
@@ -188,11 +135,126 @@ func executeStage(stage string, updateStage UpdateStage, steps []step) error {
...
+		t.RetrievalTask.v = r
+	}
+	if s != nil {
+		t.StorageTask.m = schema.Maybe_Value
+		t.StorageTask.v = s
+	}
+	return &t
+}
+
+func (t *_Task) Assign(worker string, status Status) {
```
mvdan, line 200: I feel similarly uneasy about all these extra methods, but I guess they help manipulate and use the structs since the codegen types are pretty hard to use directly.

tasks/storage_deal.go
```diff
@@ -321,3 +324,15 @@ func logStages(info api.DealInfo, log LogStatus) {
 		)
 	}
 }
+
+func (sp *_StorageTask__Prototype) Of(miner string, maxPrice, size, startOffset int64, fastRetrieval, verified bool) StorageTask {
+	st := _StorageTask{
```
mvdan, line 329: if you wanted to go "proper ipld-prime", this should use an assembler, which will do extra checks like complain if there are non-optional fields which weren't set.

And the verbosity shouldn't be too bad if you use https://pkg.go.dev/github.com/ipld/go-ipld-prime/fluent/qp#example-package.

(same applies to all the other `_T{...}` expressions)

testutil/mockbot.go
```diff
@@ -156,6 +154,10 @@ func (md *MockDaemon) worker(n int) {
 	}
 }
 
+func mustString(s string, _ error) string {
+	return s
```
mvdan, line 158: panic if err!=nil? otherwise we could be hiding pretty nasty errors.

or, if you want to ignore the error, call this something else like `tryString`.

tasks/gen/main.go
```diff
@@ -0,0 +1,128 @@
+package main
```
mvdan, line 1: it would be a bit easier to just have this in the tasks package as a `gen.go` with `// +build ignore`, since you avoid the directory. another big plus is that the gen package doesn't pollute the module in `go list ./...`.

metrics/testrecorder/testrecorder.go
```diff
@@ -24,8 +24,12 @@ func (tr *TestMetricsRecorder) Handler() http.Handler {
 	return nil
 }
 
-func (tr *TestMetricsRecorder) ObserveTask(task *tasks.Task) error {
-	tr.tasks[task.UUID] = append(tr.tasks[task.UUID], task.Status)
+func mustString(s string, _ error) string {
```
mvdan, line 27: maybe export this in the tasks package? not that the copy-pasting is significant, but it seems like an IPLD helper so it could just go in there too. And it's already been copied a few times.

controller/state/statedb.go
```diff
@@ -123,39 +122,40 @@ func (s *stateDB) db() *sql.DB {
 }
 
 // Get returns a specific task identified by ID
-func (s *stateDB) Get(ctx context.Context, taskID string) (*tasks.Task, error) {
+func (s *stateDB) Get(ctx context.Context, taskID string) (tasks.Task, error) {
 	var serialized string
```
mvdan, line 126: hm, getting serialized bytes as a string is a bit weird. I'd add a TODO about using []byte directly here.

mvdan (approved):
Overall SGTM. My only main concern is that you're using the code-genned types directly in a bunch of ways, like `_T{...}` and `t.Foo.x`, and all that adds pretty close ties to the current code generation.

If the code generation changes in any way, or if we later swap codegen for something else, some of this code will need to be re-written.

If you use the generic ipld-prime interfaces for reading and creating data, then we wouldn't have those problems. The code might get a bit more verbose, but you already need helpers like GetUUID and mustString as is, so I don't think the code would get worse per se :)

mvdan:
Also, to clarify - I agree that using composite literals and field accessors directly is nice, but it's just not a great idea with the codegen. That's one of the main reasons I'm doing the reflection experiment, to make the node API "Go type first".

warpfork:
I'm pretty fine with methods like this in the same package adding more domain-specific usability to the codegen types, fwiw.  I don't think it's terribly likely (and certainly, not going to be high frequency) for the codegen internals to change in ways that will make this a pain to maintain.

I mainly wonder if adding constructor functions like this `Of(...)` pattern on the Prototypes is something we should add to the codegen.  (I had thought about it before, and even have a half-completed diff that unfortunately didn't get carried to completion.  I'd merge proposals for it real fast if anyone wants to stab at it.)

PR #83, opened 2021-05-06
add graphql server / endpoint

willscott:
* generation of graphql schema over ipld types
* secondary servlet at separate port if specified as flag
* returns list of 'getall' tasks as initial available query.


controller/graphql/server.go
```diff
@@ -0,0 +1,96 @@
...
+				Context:       ctx,
+				Schema:        schema,
+				RequestString: r.URL.Query().Get("query"),
+			})
+		}
+
+		if len(result.Errors) > 0 {
+			log.Printf("Query had errors: %s, %v", r.URL.Query().Get("query"), result.Errors)
+		}
+		json.NewEncoder(w).Encode(result)
```
mvdan, line 92: check this error? at least we could log it and/or return a generic error string

controller/graphql/server.go
```diff
@@ -0,0 +1,96 @@
...
+
+	mux := http.NewServeMux()
+	mux.Handle("/graphql", CorsMiddleware(func(w http.ResponseWriter, r *http.Request) {
+		var result *graphql.Result
+		ctx := r.Context()
+
+		if r.Method == "POST" && r.Header.Get("Content-Type") == "application/json" {
+			var p postData
+			if err := json.NewDecoder(r.Body).Decode(&p); err != nil {
+				http.Error(w, err.Error(), 500)
```
mvdan, line 58: seems to me like this should be a BadRequest back to the user, not a ServerInternalError

controller/graphql/server.go
```diff
@@ -0,0 +1,96 @@
...
+		} else {
+			result = graphql.Do(graphql.Params{
+				Context:       ctx,
+				Schema:        schema,
+				RequestString: r.URL.Query().Get("query"),
+			})
+		}
+
+		if len(result.Errors) > 0 {
+			log.Printf("Query had errors: %s, %v", r.URL.Query().Get("query"), result.Errors)
```
mvdan, line 90: I guess the user would get these too?

mvdan (approved):

controller/graphql/server.go
willscott: yep. they're part of the result object returned

PR #89, opened 2021-05-07
update schema to account for finished tasks

willscott:
This defines what we should store about a final task, and expanded the schema of a task during execution to keep the log of previous transitioned states.

### DIFF
--- a/controller/graphql/schema.go
+++ b/controller/graphql/schema.go
@@ -36,8 +36,187 @@ func resolve_map_at(p graphql.ResolveParams) (interface{}, error) {
         return nil, fmt.Errorf("unknown key type: %T", arg)
     }
 }
-func PopTask__Status__resolve(p graphql.ResolveParams) (interface{}, error) {
-    ts, ok := p.Source.(tasks.PopTask)
+func StageDetails__Description__resolve(p graphql.ResolveParams) (interface{}, error) {
+    ts, ok := p.Source.(tasks.StageDetails)
+    if !ok {
+        return nil, errNotNode
+    }
+    
+    f := ts.FieldDescription()
+    if f.Exists() {
+        
+        return f.Must().AsString()
+        
+    } else {
+        return nil, nil
+    }
+    
+}
+func StageDetails__ExpectedDuration__resolve(p graphql.ResolveParams) (interface{}, error) {
+    ts, ok := p.Source.(tasks.StageDetails)
+    if !ok {
+        return nil, errNotNode
+    }
+    
+    f := ts.FieldExpectedDuration()
+    if f.Exists() {
+        
+        return f.Must().AsString()
+        
+    } else {
+        return nil, nil
+    }
+    
+}
+func StageDetails__Logs__resolve(p graphql.ResolveParams) (interface{}, error) {
+    ts, ok := p.Source.(tasks.StageDetails)
+    if !ok {
+        return nil, errNotNode
+    }
+    
+    return ts.FieldLogs(), nil
+    
+}
+func StageDetails__UpdatedAt__resolve(p graphql.ResolveParams) (interface{}, error) {
+    ts, ok := p.Source.(tasks.StageDetails)
+    if !ok {
+        return nil, errNotNode
+    }
+    
+    f := ts.FieldUpdatedAt()
+    if f.Exists() {
+        
+        return f.Must(), nil
+        
+    } else {
+        return nil, nil
+    }
+    
+}
+var StageDetails__type = graphql.NewObject(graphql.ObjectConfig{
+    Name: "StageDetails",
+    Fields: graphql.Fields{
+        "Description": &graphql.Field{
+            
+            Type: graphql.String,
+            
+            Resolve: StageDetails__Description__resolve,
+        },
+        "ExpectedDuration": &graphql.Field{
+            
+            Type: graphql.String,
+            
+            Resolve: StageDetails__ExpectedDuration__resolve,
+        },
+        "Logs": &graphql.Field{
+            
+            Type: graphql.NewNonNull(List_Logs__type),
+            
+            Resolve: StageDetails__Logs__resolve,
+        },
+        "UpdatedAt": &graphql.Field{
+            
+            Type: Time__type,
+            
+            Resolve: StageDetails__UpdatedAt__resolve,
+        },
+    },
+})
+var List_StageDetails__type = graphql.NewObject(graphql.ObjectConfig{
+    Name: "List_StageDetails",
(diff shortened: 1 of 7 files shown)


tasks/gen.go
```diff
@@ -105,12 +105,27 @@ func main() {
...
+		schema.SpawnStructField("RetrievalTask", "RetrievalTask", true, false),
+		schema.SpawnStructField("StorageTask", "StorageTask", true, false),
+		schema.SpawnStructField("DealID", "Int", false, false),
+		schema.SpawnStructField("MinerMultiAddr", "String", false, false),
+		schema.SpawnStructField("ClientApparentAddr", "String", false, false),
+		schema.SpawnStructField("MinerLatencyMS", "Int", true, false),
+		schema.SpawnStructField("TimeToFirstByteMS", "Int", true, false),
+		schema.SpawnStructField("TimeToLastByteMS", "Int", true, false),
+		schema.SpawnStructField("Events", "Link_List_StageDetails", false, false),
+	}, schema.SpawnStructRepresentationMap(map[string]string{})))
```
mvdan, line 127: you can just use `nil` in these map arguments, FYI

tasks/gen.go
```diff
@@ -105,12 +105,27 @@ func main() {
 		schema.SpawnStructField("WorkedBy", "String", true, false),
 		schema.SpawnStructField("Stage", "String", false, false),
 		schema.SpawnStructField("CurrentStageDetails", "StageDetails", true, false),
+		schema.SpawnStructField("PastStageDetails", "List_StageDetails", true, false),
```
mvdan, line 108: @gammazero can answer this better, but I believe we do have this data already. It's just separate entries in the database, since it's like a journal. We have a unique constraint on:

> task_status_ledger.uuid, task_status_ledger.status, task_status_ledger.stage

If we only stored the last state, this constraint wouldn't make sense - we would simply have unique tasks by UUID.

mvdan (commented):
SGTM, but I'd wait for @gammazero because I think we're duplicating data.

gammazero (approved):

controller/graphql/schema.go
```diff
@@ -460,8 +588,58 @@ var union__Any__Link = graphql.NewObject(graphql.ObjectConfig{
     },
 })
 
-var Tasks__type = graphql.NewObject(graphql.ObjectConfig{
-    Name: "Tasks",
+func RetrievalTask__Miner__resolve(p graphql.ResolveParams) (interface{}, error) {
+    ts, ok := p.Source.(tasks.RetrievalTask)
+    if !ok {
+        return nil, errNotNode
```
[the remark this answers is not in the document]
mvdan: you're reviewing generated code :)

tasks/ipldsch_satisfaction.go
```diff
@@ -1628,145 +1655,326 @@ func (m MaybeFloat) AsNode() ipld.Node {
...
+	na.ca_TimeToFirstByteMS.reset()
+	na.ca_TimeToLastByteMS.reset()
+	na.ca_Events.reset()
 }
 
-func (na *_Float__Assembler) reset() {}
-func (_Float__Assembler) BeginMap(sizeHint int64) (ipld.MapAssembler, error) {
-	return mixins.FloatAssembler{"tasks.Float"}.BeginMap(0)
+var (
+	fieldBit__FinishedTask_Status = 1 << 0
```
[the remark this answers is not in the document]
mvdan: ditto generated code. if you feel strongly about this, raise an issue in go-ipld-prime for its codegen.

tasks/ipldsch_satisfaction.go
```diff
@@ -1777,49 +1985,2232 @@ func (na *_Float__Assembler) AssignNode(v ipld.Node) error {
...
+		ma.state = maState_midValue
+		ma.f = 6
+		ma.ca_ClientApparentAddr.w = &ma.w.ClientApparentAddr
+		ma.ca_ClientApparentAddr.m = &ma.cm
+		return &ma.ca_ClientApparentAddr, nil
+	case "MinerLatencyMS":
+		if ma.s & fieldBit__FinishedTask_MinerLatencyMS != 0 {
+			return nil, ipld.ErrRepeatedMapKey{&fieldName__FinishedTask_MinerLatencyMS}
+		}
+		ma.s += fieldBit__FinishedTask_MinerLatencyMS
```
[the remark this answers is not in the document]
mvdan: ditto generated code. if you feel strongly about this, raise an issue in go-ipld-prime for its codegen.

willscott:
Updated this now "probably too large" PR to
* use the ipld linksystem to calculate CIDs for tasks whenever they are saved
* Calculate the schema-defined `finalizedTask` item when a task moves to a success or failed state.
* The remaining work we need to do here is fill out [This https://github.com/filecoin-project/dealbot/pull/89/files#diff-db62cd37a290e83d0c821b86427dc6c192ec971801bc690a8184a037e830b262R295] - how do we actually calculate e.g. the latency, and deal that was made from the various logs recorded by the bot. Some will be specific logs, while others will be time deltas between two different logs.

tasks/model.go
```diff
@@ -92,6 +102,14 @@ var RetrievalStages = map[string]StageDetails{
 	"DealComplete":      asStageDetails("All bytes received and deal is completed", "a few seconds"),
 }
 
+// the multi-codec and hash we use for cid links by default
+var linkProto = linksystem.LinkBuilder{cid.Prefix{
+	Version:  1,
+	Codec:    0x0129, //dagjson
+	MhType:   0x12,   // sha2-256
```
mvdan, line 109: ditto

tasks/model.go
```diff
@@ -197,42 +215,102 @@ func (tp *_Task__Prototype) New(r RetrievalTask, s StorageTask) Task {
...
 }
 
-func (t *_Task) UpdateTask(tsk UpdateTask) error {
-	t.Status = tsk.Status
-	if tsk.CurrentStageDetails.Exists() {
-		t.CurrentStageDetails = tsk.CurrentStageDetails
-	} else {
-		t.CurrentStageDetails = _StageDetails__Maybe{m: schema.Maybe_Absent}
+func (t *_Task) Finalize(ctx context.Context, s ipld.Storer) (FinishedTask, error) {
+	if t.Status != *Failed && t.Status != *Successful {
```
mvdan, line 266: I'm a bit uneasy that we always dereference these globals when using them. Maybe they shouldn't be pointers.

willscott: i think this is a consequence of how ipld deals with exported types (the 'Status' vs '_Status' thing)

PR #93, opened 2021-05-07
don't leak devnet in the new integration test

mvdan:
(see commit message)


mvdan:
FYI @gammazero, this is what we talked about.

I think this should fix CI too. I stopped using the fork to kick off Circle.

willscott (approved):

mvdan:
I've disabled the test on CI for now, to unblock main.

PR #102, opened 2021-05-11
Support multiple concurrent workers

gammazero:
- Configurable number of workers, default 1
- Engine.Close() cancels workers
- Each task has its own lifetime timeout

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -16,11 +16,10 @@ import (
 )
 
 const (
-	nextTaskTimeout = 10 * time.Second
-	noTasksWait     = 5 * time.Second
-	maxTaskLifetime = 24 * time.Hour
-
 	defaultWorkers = 1
+
+	maxTaskLifetime = 24 * time.Hour
+	noTasksWait     = 5 * time.Second
 )
 
 var log = logging.Logger("engine")
@@ -65,6 +64,7 @@ func New(ctx context.Context, cliCtx *cli.Context) (*Engine, error) {
 		host:       uuid.New().String()[:8], // TODO: set from config toml
 	}
 
+	// Create context to cancel workers on Close
 	ctx, e.cancel = context.WithCancel(ctx)
 
 	e.wg.Add(workers)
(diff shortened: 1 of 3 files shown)


engine/engine.go
```diff
@@ -14,6 +15,13 @@ import (
 	logging "github.com/ipfs/go-log/v2"
 )
 
+const (
+	defaultWorkers = 1
```
mvdan, line 19: join this with the `1` constant in the flag?

engine/engine.go
```diff
@@ -50,82 +64,134 @@ func New(ctx context.Context, cliCtx *cli.Context) (*Engine, error) {
...
+	e.wg.Wait() // wait for workers to stop
 	e.closer()
 }
 
-func (e *Engine) worker(n int) {
-	log.Infow("engine worker started", "worker_id", n)
-
+func (e *Engine) popTask(ctx context.Context) tasks.Task {
+	// Pop tasks until a task is found or until no more tasks
 	for {
```
mvdan, line 86: why is this in a loop? the popTask caller already does a loop with 5s waits. if you remove the loop, then the ctx.Err check should also go.

engine/engine.go
```diff
@@ -50,82 +64,134 @@ func New(ctx context.Context, cliCtx *cli.Context) (*Engine, error) {
...
+			select {
+			case <-ctx.Done():
+				return // engine closed, worker exits.
+			case <-time.After(noTasksWait):
 			}
+			continue
 		}
-		_, err = e.client.UpdateTask(ctx, task.UUID.String(),
+
+		log.Infow("successfully acquired task", "uuid", task.UUID)
```
mvdan, line 125: log the "worker ID" here too

engine/engine.go
```diff
@@ -50,82 +64,134 @@ func New(ctx context.Context, cliCtx *cli.Context) (*Engine, error) {
...
+				log.Warn("task", task.UUID.String(), "canceled for shutdown")
+				return
+			}
+			finalStatus = tasks.Failed
+			log.Errorw("retrieval task returned error", "err", err)
+		} else {
+			log.Info("successfully retrieved data")
+		}
+	}
+	if task.StorageTask.Exists() {
```
mvdan, line 166: else if? or error if both exist

mvdan (approved):

engine/engine.go
gammazero: Done

engine/engine.go
gammazero: This is in a loop to skip over the unusable tasks.  Probably not really necessary, so removed.

engine/engine.go
gammazero: Removed this, flag already has default.

engine/engine.go
gammazero: Copied from original logic in engine, so I thought this may have been in case some task could have both types.
https://github.com/filecoin-project/dealbot/blob/main/engine/engine.go#L110

Changing to `else if`

willscott (approved):

PR #103, opened 2021-05-12
all: make devnet and integration tests more reliable

mvdan:
(see commit message)

### DIFF
--- a/devnet/devnet.go
+++ b/devnet/devnet.go
@@ -6,54 +6,62 @@ import (
 	"os"
 	"os/exec"
 	"os/signal"
+	"path/filepath"
 	"strings"
 	"sync"
 	"syscall"
 	"time"
 )
 
-func runShell(ctx context.Context, name string, commands []string) {
+func runCmdsWithLog(ctx context.Context, name string, commands [][]string) {
 	logFile, err := os.Create(name + ".log")
 	if err != nil {
 		log.Fatal(err)
 	}
 	defer logFile.Close()
 
-	src := strings.Join(commands, " && ")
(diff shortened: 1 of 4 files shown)


mvdan:
This took a significant amount of time since waiting for Circle+devnet is a slow iteration process :) But we should be all good now!

devnet/devnet.go
```diff
@@ -6,71 +6,77 @@ import (
...
-func runMiner(ctx context.Context) {
-	time.Sleep(5 * time.Second) // wait for lotus node to run
-
-	lotusMinerCmds := []string{
-		"lotus wallet import ~/.genesis-sectors/pre-seal-t01000.key",
-		"lotus-miner init --genesis-miner --actor=t01000 --sector-size=2048 --pre-sealed-sectors=~/.genesis-sectors --pre-sealed-metadata=~/.genesis-sectors/pre-seal-t01000.json --nosync",
-		"lotus-miner run --nosync",
+func runLotusMiner(ctx context.Context, home string) {
+	cmds := [][]string{
+		{"lotus", "wait-api"}, // wait for lotus node to run
```
mvdan, line 51: note that I also replaced the 5s sleep with this command, following what's done in header.sh.

devnet/devnet.go
```diff
@@ -85,6 +91,7 @@ func setDefaultWalletCmd(ctx context.Context) {
 		if err != nil {
 			continue
 		}
+		// TODO: stop once we've set the default wallet once.
```
mvdan, line 94: setting the default wallet could be faster and less hacky, but it doesn't feel urgent right now. I left TODOs.

willscott (approved):

PR #106, opened 2021-05-12
add command for draining a worker

willscott:
makes a table, with a handler to indicate tasks shouldn't go to a specific bot, and querying on assignments.

### DIFF
--- a/controller/state/migrations/000003_initialize_schema.down.sql
+++ b/controller/state/migrations/000003_initialize_schema.down.sql
@@ -1,5 +1,5 @@
 BEGIN;
 
-DROP TABLE IF EXISTS drainedWorkers;
+DROP TABLE IF EXISTS drained_workers;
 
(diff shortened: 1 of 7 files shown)


controller/state/migrations/000003_initialize_schema.down.sql
```diff
@@ -0,0 +1,5 @@
+BEGIN;
+
+DROP TABLE IF EXISTS drainedWorkers;
```
gammazero, line 3: To be consistent with other table names, use: `drained_workers`

... although I just realized we have `finalizedData`.

gammazero (approved):
Looks good. Minor issue with table naming convention.

Do we need anything to remove entries from the `drainedWorkers` table?

willscott:
i think once we go to rotate a few times we'll have a much better sense of expiring / removing draining

PR #108, opened 2021-05-13
add structure for record updates

willscott:
this builds the top level scaffold expected to be consumed by reputation systems

* A new command for now to indicate all work by a given bot is done, at which point all finalized tasks from that bot are collected into a record update that is set as the new 'head'
* A graphql endpoint to query against this chain.

### DIFF
--- a/controller/state/record_update.go
+++ b/controller/state/record_update.go
@@ -1,0 +1,10 @@
+package state
+
+type record_update_status int
+
+const (
+	LATEST_UPDATE     record_update_status = 1
+	PREVIOUS_UPDATE   record_update_status = 2
+	UNATTACHED_RECORD record_update_status = 10
+	ATTACHED_RECORD   record_update_status = 11
+)
--- a/controller/state/record_update.go
+++ b/controller/state/record_update.go
@@ -1,10 +1,10 @@
 package state
 
-type record_update_status int
+type RecordUpdateStatus int
 
 const (
-	LATEST_UPDATE     record_update_status = 1
-	PREVIOUS_UPDATE   record_update_status = 2
-	UNATTACHED_RECORD record_update_status = 10
-	ATTACHED_RECORD   record_update_status = 11
+	LATEST_UPDATE     RecordUpdateStatus = 1
+	PREVIOUS_UPDATE   RecordUpdateStatus = 2
+	UNATTACHED_RECORD RecordUpdateStatus = 10
+	ATTACHED_RECORD   RecordUpdateStatus = 11
 )
--- a/controller/state/statedb.go
+++ b/controller/state/statedb.go
@@ -357,7 +357,11 @@ func (s *stateDB) Update(ctx context.Context, taskID string, req tasks.UpdateTas
 				return err
 			}
 
-			if _, err = tx.ExecContext(ctx, addHeadSQL, flink.(linksystem.Link).Cid.String(), time.Now(), req.WorkedBy.String(), UNATTACHED_RECORD); err != nil {
+			if _, err := tx.ExecContext(ctx, addHeadSQL,
+				flink.(linksystem.Link).Cid.String(),
+				time.Now(),
+				req.WorkedBy.String(),
+				UNATTACHED_RECORD); err != nil {
 				return err
 			}
 		}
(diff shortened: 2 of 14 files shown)


controller/state/record_update.go
```diff
@@ -0,0 +1,10 @@
+package state
+
+type record_update_status int
+
+const (
+	LATEST_UPDATE     record_update_status = 1
+	PREVIOUS_UPDATE   record_update_status = 2
+	UNATTACHED_RECORD record_update_status = 10
+	ATTACHED_RECORD   record_update_status = 11
```
mvdan, line 9: interesting choice of casing, for Go :)

controller/state/statedb.go
```diff
@@ -348,10 +348,18 @@ func (s *stateDB) Update(ctx context.Context, taskID string, req tasks.UpdateTas
...
+			finalized, err := updatedTask.Finalize(ctx, txContextStorer(ctx, tx))
 			if err != nil {
 				return err
 			}
+			flink, err := linkProto.Build(ctx, ipld.LinkContext{}, finalized, txContextStorer(ctx, tx))
+			if err != nil {
+				return err
+			}
+
+			if _, err = tx.ExecContext(ctx, addHeadSQL, flink.(linksystem.Link).Cid.String(), time.Now(), req.WorkedBy.String(), UNATTACHED_RECORD); err != nil {
```
mvdan, line 360: might as well use `:=` here

mvdan: and perhaps split the line as it's quite long

controller/state/statedb.go
```diff
@@ -483,6 +501,109 @@ func (s *stateDB) countTasks(ctx context.Context) (int, error) {
...
+	}
+
+	rcrds := []tasks.AuthenticatedRecord{}
+
+	for itms.Next() {
+		var lnk string
+		if err := itms.Scan(&lnk); err != nil {
+			return err
+		}
+		if _, err = tx.ExecContext(ctx, updateHeadSQL, ATTACHED_RECORD, lnk); err != nil {
```
mvdan, line 573: ditto `:=`

controller/state/statedb.go
```diff
@@ -483,6 +501,109 @@ func (s *stateDB) countTasks(ctx context.Context) (int, error) {
...
+			return err
+		}
+		c, err := cid.Decode(lnk)
+		if err != nil {
+			return err
+		}
+		rcrd := tasks.Type.AuthenticatedRecord.Of(c, sig)
+		rcrds = append(rcrds, rcrd)
+	}
+	rcrdlst := tasks.Type.List_AuthenticatedRecord.Of(rcrds)
```
mvdan, line 587: reading these names is making my head hurt a little... how about just `records` and `recordList`?

tasks/gen.go
```diff
@@ -138,6 +139,18 @@ func main() {
...
+	// reputation api
+	ts.Accumulate(schema.SpawnStruct("AuthenticatedRecord", []schema.StructField{
+		schema.SpawnStructField("Record", "Link_FinishedTask", false, false),
+		schema.SpawnStructField("Signature", "Bytes", false, false),
+	}, schema.SpawnStructRepresentationMap(map[string]string{})))
+	ts.Accumulate(schema.SpawnList("List_AuthenticatedRecord", "AuthenticatedRecord", false))
+	ts.Accumulate(schema.SpawnStruct("RecordUpdate", []schema.StructField{
+		schema.SpawnStructField("Records", "List_AuthenticatedRecord", false, false),
+		schema.SpawnStructField("SigPrev", "Bytes", false, false),
+		schema.SpawnStructField("Previous", "Link", false, false),
```
mvdan, line 151: Link to...? I imagine this could/shoudl be something like Link_RecordUpdate.

mvdan (approved):
Maybe document somewhere why you're using a chain-like structure with signatures and a "head". I can guess why we'd want that, but it's not entirely obvious.

controller/state/record_update.go
willscott: thoughts on preferred case for these constants? jut `LatestUpdate`?

controller/state/statedb.go
willscott: in the reputation system parlance we call `Task` a `Record`, and then have a surrounding `AuthenticatedRecord` that is a signature over the cid of the record.

tasks/gen.go
willscott: this seems to be a bug in graphql code gen. If i have a cycle of links (struct includes link to itself) we end up with a circular definition.

PR #109, opened 2021-05-13
fix logging of tasks since we moved to IPLD

mvdan:
(see commit message)

### DIFF
--- a/metrics/log/log.go
+++ b/metrics/log/log.go
@@ -8,7 +8,6 @@ import (
 	"github.com/filecoin-project/dealbot/metrics"
 	"github.com/filecoin-project/dealbot/tasks"
 	logging "github.com/ipfs/go-log/v2"
-	"github.com/ipld/go-ipld-prime"
 )
 
 type logRecorder struct {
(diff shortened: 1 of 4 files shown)


metrics/log/log.go
```diff
@@ -23,8 +24,23 @@ func (lr *logRecorder) Handler() http.Handler {
...
 }
 
-func mustString(s string, _ error) string {
-	return s
+func simpleGoValue(node ipld.Node) interface{} {
+	var val interface{}
+	var err error
+	switch kind := node.Kind(); kind {
+	case ipld.Kind_String:
+		val, err = node.AsString()
```
willscott, line 32: can these be simplified to `return node.String()`

willscott (approved):

metrics/log/log.go
mvdan: hm, I'm pretty sure that's not an IPLD method. it looks like we define it for Status, but that's for an Int that we want to behave like a human-friendly string.

if that's something we want to do for logging (the answer is probably "yes"), then for Status fields we could call the String method instead of this func here.

And I'm actually just realising that its String method might not work in practice. It uses a pointer-based map, so if we do an IPLD codec decode, we'd end up with different pointers and... no useful string :)

willscott: I've been doing that through a bunch of this code and it seems to work, e.g 
https://github.com/filecoin-project/dealbot/blob/main/controller/state/statedb.go#L492

PR #119, opened 2021-05-18
Tasks can be scheduled for repeated execution 

gammazero:
Storage and retrieval tasks have a "Schedule" string, which when not empty specifies a schedule for repeated task execution.  The tasks also have an optional  "ScheduleLimit" string which is a golang duration string (e.g. "72h3m") and specifies for how long the task will keep getting scheduled.  If a task has no schedule then it is run immediately, one time. If a task has a Schedule, but no ScheduleLimit, then the task is scheduled indefinitely.

Each time a task is executed, it is started from the same state as when it was retrieved by PopTask.

Draft until longer-term test can be performet.


scheduler/job.go
```diff
@@ -0,0 +1,69 @@
...
+	"context"
+	"sync/atomic"
+	"time"
+
+	"github.com/filecoin-project/dealbot/tasks"
+	"github.com/robfig/cron/v3"
+)
+
+// Job is a set of parameters for running a job
+type Job struct {
```
willscott, line 13: having 2 contexts on this struct feels awkward.

willscott (approved):

scheduler/job.go
gammazero: Having one context in a struct is bad enough:
> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it.

-- https://golang.org/pkg/context/

So, two cancels out, right?  Ok, maybe not so much.  I originally had put that together using channels, and it seemed even more awkward because of trying to use the context passed into the engine constructor.  Maybe the answer is to just ditch that context (use ctx for cancelation, to for lifecycle management) and this will all clear up.  I will take another pass at that.

engine/engine.go
```diff
@@ -191,3 +229,28 @@ func (e *Engine) runTask(ctx context.Context, task tasks.Task) {
...
 	}
 }
+
+func getTaskSchedule(task tasks.Task) (string, time.Duration) {
+	var schedule, limit string
+	var duration time.Duration
+
+	if task.RetrievalTask.Exists() {
+		schedule = task.RetrievalTask.Must().Schedule.Must().String()
+		limit = task.StorageTask.Must().ScheduleLimit.Must().String()
```
mvdan, line 239: did you mean RetrievalTask here? I think you would avoid mistakes like these with...

```
if t := task.RetrievalTask; t.Exists {
    schedule = t.Must()...
```

scheduler/job.go
```diff
@@ -0,0 +1,69 @@
...
+
+// Send job to worker, wait for worker to signal it is finished.
+func (j *Job) Run() {
+	// If the schedule has expired remove job from scheduler.
+	if !j.expireAt.IsZero() && time.Now().After(j.expireAt) {
+		j.sched.Remove(j.entryID)
+		return
+	}
+
+	atomic.AddInt32(&j.run, 1)
```
mvdan, line 35: I'm surprised that atomic is necessary; I'd assume that it would be impossible for a cron job to fire twice at the same time.

controller/state/interface.go
```diff
@@ -12,7 +12,7 @@ type State interface {
 	Get(ctx context.Context, uuid string) (tasks.Task, error)
 	GetAll(ctx context.Context) ([]tasks.Task, error)
 	GetHead(ctx context.Context) (tasks.RecordUpdate, error)
-	Update(ctx context.Context, uuid string, req tasks.UpdateTask) (tasks.Task, error)
+	Update(ctx context.Context, uuid string, req tasks.UpdateTask, run int) (tasks.Task, error)
```
mvdan, line 15: The name "run" doesn't really tell me much if I'm just reading this interface. Perhaps `runCounter`? Another option is to start writing docs for the non-obvious parameters.

Also, did you consider just adding this to UpdateTask?

scheduler/scheduler.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+// RunChan returns a channel that sends jobs to be run
+func (s *Scheduler) RunChan() <-chan *Job {
+	return s.runChan
+}
+
+// Add adds a Task to be scheduled.
+func (s *Scheduler) Add(cronExp string, task tasks.Task, maxRunTime, scheduleLife time.Duration) (cron.EntryID, error) {
+	job := &Job{
```
mvdan, line 60: If I understand correctly, a cron job can only be run as long as the controller is up, because if it gets shut down and re-started, the atomic `run` counter is lost. We'd start running the cron job from counter 1, which would presumably make the DB unhappy.

scheduler/scheduler_test.go
```diff
@@ -0,0 +1,167 @@
...
+func TestScheduleTask(t *testing.T) {
+	s := NewWithSeconds()
+
+	// Add task to scheduler
+	jid, err := s.Add(everySecond, nil, maxRunTime, 0)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	var noEnt cron.Entry
```
mvdan, line 25: you could make this a global like zeroEntry :)

scheduler/scheduler_test.go
```diff
@@ -0,0 +1,167 @@
...
+		t.Errorf("job has wrong run count, expected 2, got %d", job.RunCount())
+	}
+
+	job.End()
+
+	// Test that job stops immediately
+	select {
+	case <-job.RunContext().Done():
+		t.Log("stopped", job.ID(), "run", job.RunCount())
+	case <-time.After(time.Millisecond):
```
mvdan, line 76: I think these `1ms` timeouts might give us trouble, especially in CI where machines are often shared with other concurrent jobs. Something like 100ms or 200ms would be much safer.

scheduler/job.go
gammazero: It was because the race detector complained about a race here and where this is read.  Still should not be needed, because is not read until the job is read from the channel.

mvdan: Hm, if the race detector complains when we think it shouldn't, then it smells like something is wrong. Why would we ever read and write to this field at the same time? It would be extremely surprising for the race detector to be buggy :)

controller/state/interface.go
gammazero: Will rename to `runCount`.  I was not sure if keeping a run count was OK or if a new task was needed, and I for some reason thought this would be easier to change.  It really should be in UpdateTask.

engine/engine.go
gammazero: Much nicer.  Also, obviously, I did not test the retrieval task with a limit.

scheduler/scheduler.go
gammazero: Yes, this needs to have the run counter to start at.

Also, I need to add code to resume scheduling tasks there were already running.  These include:
1. Tasks that are completed, but should be run periodically according to their schedule
2. Tasks that were running, but were abandoned during some stage of execution.

I plan to do this in a follow-up PR. That will handle the recovery of running tasks and rescheduling of scheduled tasks.

Currently, once a task has been popped from the available tasks, it will not get popped off again.  When dealbot is restarted it will not resume any tasks that were already started, until that PR.

scheduler/job.go
gammazero: OK, fixed this by copying a subset of the job info over the channel.

scheduler/job.go
gammazero: Got rid of all contexts in structs.  Ends up being a much nicer solution.

scheduler/scheduler_test.go
gammazero: Done

scheduler/scheduler_test.go
gammazero: Done.  This particular wait is not even needed.

willscott:
is this ready to come out of draft?

gammazero:
@willscott This can come out of draft and appears to be working on the test network.

willscott (approved):

gammazero:
@willscott Fixed now.  Should be good to merge

PR #120, opened 2021-05-18
run lotus fetch-params explicitly

mvdan:
(see commit message)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,16 +16,12 @@ commands:
     steps:
       - run: sudo apt-get update
       - run: sudo apt-get install ocl-icd-opencl-dev libhwloc-dev
-      - run: git clone https://github.com/filecoin-project/lotus.git /tmp/lotus
-      - run: cd /tmp/lotus && git submodule sync && git submodule update --init
+      - run: git clone --depth=1 https://github.com/filecoin-project/lotus /tmp/lotus
+      - run: cd /tmp/lotus && git submodule update --init --depth=1
       - run: cd /tmp/lotus && make debug && sudo make install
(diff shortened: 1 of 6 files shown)


willscott:
I'm worried that once we do this we are going to have to have a pretty constant stream of updates that we don't have automation on in order to keep the lotus we're testing against up to date. There's a pretty rapid release schedule, and in order to test deal successes, we want the bot to always be against the latest version. I wonder if we have a way to know which version has been promoted to infra and automatically have our tests also run against that?

mvdan:
You're right, though always using the latest master has the same problem and also makes CI/testing more unreliable.

How often does infra get its Lotus version updated, and how? In that process, we could possibly introduce a `sed` find-and-replace command to update the version picked for testing, too. cc @coryschwartz

gammazero (approved):
LGTM

mvdan:
@willscott what do you want to do here, then? We can agree that pinning Lotus in separate ways for testing and deploying is not great, but not pinning Lotus at all for testing seems worse however you look at it :) Especially since we're currently following master, which could break in unexpected ways at any moment.

I'm happy to look at ways to make the testing and deploying pinning to always be in sync, but I'll need some collaboration with infra on that.

willscott:
I'd rather see tests break and realize it's lotus as a heads up, rather than having that happen first in an infra deployment?
Lotus has a fairly high bar before they merge to master, so i don't think we run into lotus breaking, and if things have been accepted/merged that break this, that's probably a thing we want to know through tests at that point and be able to make noise about it.

mvdan:
@willscott I've undone the bits about pinning lotus, as I think the other changes are still worth merging :)

willscott (approved):

PR #122, opened 2021-05-19
add /Car endpoint

willscott:
for getting the ipld dag of finalized tasks

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -121,6 +121,7 @@ func NewWithDependencies(listener, graphqlListener net.Listener, recorder metric
 	r.HandleFunc("/status", srv.reportStatusHandler).Methods("POST")
 	r.HandleFunc("/tasks/{uuid}", srv.updateTaskHandler).Methods("PATCH")
 	r.HandleFunc("/tasks/{uuid}", srv.getTaskHandler).Methods("GET")
+	r.HandleFunc("/car", srv.carHandler).Methods("GET")
 	r.Methods("OPTIONS").HandlerFunc(srv.sendCORSHeaders)
 	metricsHandler := recorder.Handler()
 	if metricsHandler != nil {
--- a/controller/state/statedb.go
+++ b/controller/state/statedb.go
@@ -8,12 +8,14 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net/http"
 	"sync"
 	"time"
 
 	"github.com/filecoin-project/dealbot/metrics"
 	"github.com/filecoin-project/dealbot/tasks"
+	blockformat "github.com/ipfs/go-block-format"
 	"github.com/ipfs/go-cid"
 	logging "github.com/ipfs/go-log/v2"
 	"github.com/ipld/go-ipld-prime"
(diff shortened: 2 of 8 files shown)


controller/state/statedb.go
```diff
@@ -8,12 +8,14 @@ import (
...
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net/http"
 	"sync"
 	"time"
 
 	"github.com/filecoin-project/dealbot/metrics"
 	"github.com/filecoin-project/dealbot/tasks"
+	blockformat "github.com/ipfs/go-block-format"
```
hannahhoward, line 18: non-blocking: for better or worse, in most repos the shortname used here is 'blocks'

controller/tasks.go
```diff
@@ -241,6 +245,40 @@ func (c *Controller) getTaskHandler(w http.ResponseWriter, r *http.Request) {
...
+		log.Errorw("get task DB error", "err", err.Error())
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
+
+	w.WriteHeader(http.StatusOK)
+
+	ssb := builder.NewSelectorSpecBuilder(basicnode.Prototype__Any{})
+	ss := ssb.ExploreRecursive(selector.RecursionLimitNone(), ssb.ExploreUnion(
+		ssb.Matcher(),
```
hannahhoward, line 268: Non-blocking: FYI, for the purposes of CAR writing, the ExploreUnion and the Matcher are unnecessary

controller/tasks.go
```diff
@@ -241,6 +245,40 @@ func (c *Controller) getTaskHandler(w http.ResponseWriter, r *http.Request) {
...
+		ssb.ExploreAll(ssb.ExploreRecursiveEdge()),
+	))
+	root := car.Dag{
+		Root:     rootCid,
+		Selector: ss.Node(),
+	}
+	sc := car.NewSelectiveCar(r.Context(), store, []car.Dag{root})
```
hannahhoward, line 275: side note: we should make a version of selective car that actually just takes an IPLD loader -- under the hood, this is actually converting the reader BACK to a loader, which is funny case you converted a loader to a reader to get this to work :)

controller/controller.go
```diff
@@ -121,6 +121,7 @@ func NewWithDependencies(listener, graphqlListener net.Listener, recorder metric
 	r.HandleFunc("/status", srv.reportStatusHandler).Methods("POST")
 	r.HandleFunc("/tasks/{uuid}", srv.updateTaskHandler).Methods("PATCH")
 	r.HandleFunc("/tasks/{uuid}", srv.getTaskHandler).Methods("GET")
+	r.HandleFunc("/car", srv.carHandler).Methods("GET")
```
hannahhoward, line 124: This strikes me as being worthy of an HTTP test in http_test.go :)

hannahhoward (approved):
Would be great to add a test of the endpoint in the HTTP test file (vaguely assuming we can populate the database somehow?)

Not gonna block on it but consider it a strong recommend? At least for a seperate ticket?

willscott:
Added at test. Note we're now on a weird offshoot ipld prime commit i made to backport functional `dag-json` to the v0.7 we're stuck at

Issue #125, opened 2021-05-19
don't see expected 'paststagedetails' on tasks


willscott:
when looking at tasks in success/failure state, they don't seem to have paststagedetails set at all. when looking at an in-progress task, i see much less of a log than i'm expecting. are we not properly round-tripping / appending to that log?

mvdan:
I was looking into #88 and was also stumped by this. I didn't spend much time digging into it yet though.

mvdan:
I'm looking into this further. The stages are there, and they get appended; here's what I see with a few debug logs (don't mind the broken timestamps, that's on me):

```
$ grep 'added stage' dealbot-controller.log 
added stage: 51383757511-07-19 02:19:06 +0100 BST "Miner is online"
added stage: 51383757512-01-02 20:35:41 +0000 GMT "Miner responds to query ask"
added stage: 51383757512-03-21 06:44:49 +0000 GMT "Miner meets price criteria"
added stage: 51383757512-05-08 23:54:17 +0100 BST "Importing data into Lotus"
added stage: 51383757513-01-08 21:08:09 +0000 GMT "Reserving client funds"
added stage: 51383760130-08-12 13:39:23 +0100 BST "Checking for deal acceptance"
```

But when finalized, I don't see the same past stages. Debugging some more.

mvdan:
It gets weirder. When we fetch the task from the database, PastStageDetails is never there in the raw dagjson. And when we store the dagjson back, there's just one past stage in the list - which at least makes sense, since we append one element to an empty list.

And when we go to finalize the task, PastStageDetails is absent, just like it is in all the Update queries.

This doesn't seem like an IPLD problem to me, because looking at the dagjson bytes, it looks good. And the append also seems to be working well. So it seems to be the database layer that's making us forget the PastStageDetails list.

I've pushed my debugging code to https://github.com/filecoin-project/dealbot/tree/debug-past-stages, and I'm attaching dealbot-controller.log from a run of the second integration test on that branch.
[dealbot-controller.log](https://github.com/filecoin-project/dealbot/files/6516155/dealbot-controller.log)

@gammazero could you please help me debug this? You know the SQL code better than me, and currently I'm a bit stuck. I can also do some pair programming/debugging in an hour or so if you're up for it.

gammazero:
Confirmed every status/state update is being written to DB.  When the final status shows up I query task history to log all items.  However, it looks like there is a failure to read task history:
```
[...]

PR #129, opened 2021-05-20
(**WIP**) Crude tool to parse success/failure fields from .car controller export

mgoelzer:
This is a very simple tool to read an export `.car` file from `http://localhost:controller_port/car` and print its contents out in a format useful for downstream consumers.

Error message inclusion is blocked on https://github.com/filecoin-project/dealbot/issues/125

Needs ability to read directly from `http://localhost:controller_port/tasks` instead of going through the car export process.


util/read-car.go
```diff
@@ -0,0 +1,98 @@
...
+		retrievalTasksMap := dealbotJsonDataMap["RetrievalTask"].(map[string]interface{})
+		payloadCid := retrievalTasksMap["PayloadCID"]
+		miner := retrievalTasksMap["Miner"]
+
+		// final output
+		fmt.Printf("status = %s\n",statusStr)
+		fmt.Printf("datetime = %s\n", startDatetime.String())
+		fmt.Printf("ipfs location = https://ipfs.io/ipfs/%s\n",payloadCid)
+		fmt.Printf("miner = %s\n",miner)
+		fmt.Printf("\n")
```
willscott, line 96: is this the format you want this? (vs json?)

mgoelzer: It's just to prove that the fields we need can be printed out.  You're right that they are not in the correct json format yet.  That's not hard and I do plan to do it.

mgoelzer:
Closing for now.  May resurrect at some point.

PR #135, opened 2021-05-21
Implement tag matching for tasks

gammazero:
- Tasks may have a tag that designates them for certain dealbots.
- Dealbots may be configured with 0 or more tags to request tasks by.
- Tasks will to tag match any request
- Requests that specify no tags can get any task

### DIFF
--- a/commands/flags.go
+++ b/commands/flags.go
@@ -116,11 +116,10 @@ var DaemonFlags = append(DealFlags, append(CommonFlags, append(EndpointFlags, []
 		EnvVars: []string{"DEALBOT_WORKERS"},
 		Value:   1,
 	}),
-	altsrc.NewIntFlag(&cli.IntFlag{
+	altsrc.NewStringSliceFlag(&cli.StringSliceFlag{
 		Name:    "tags",
 		Usage:   "comma separated tag strings",
 		EnvVars: []string{"DEALBOT_TAGS"},
-		Value:   1,
 	}),
 }...)...)...)
 
--- a/commands/flags.go
+++ b/commands/flags.go
@@ -116,6 +116,12 @@ var DaemonFlags = append(DealFlags, append(CommonFlags, append(EndpointFlags, []
 		EnvVars: []string{"DEALBOT_WORKERS"},
 		Value:   1,
 	}),
+	altsrc.NewIntFlag(&cli.IntFlag{
+		Name:    "tags",
+		Usage:   "comma separated tag strings",
+		EnvVars: []string{"DEALBOT_TAGS"},
+		Value:   1,
+	}),
 }...)...)...)
 
 var MockFlags = []cli.Flag{
--- a/controller/state/statedb.go
+++ b/controller/state/statedb.go
@@ -10,6 +10,7 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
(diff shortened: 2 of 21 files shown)


commands/flags.go
```diff
@@ -116,6 +116,12 @@ var DaemonFlags = append(DealFlags, append(CommonFlags, append(EndpointFlags, []
 		EnvVars: []string{"DEALBOT_WORKERS"},
 		Value:   1,
 	}),
+	altsrc.NewIntFlag(&cli.IntFlag{
+		Name:    "tags",
+		Usage:   "comma separated tag strings",
+		EnvVars: []string{"DEALBOT_TAGS"},
+		Value:   1,
```
willscott, line 123: should this be a string array flag?

willscott (commented):
looks reasonable

commands/flags.go
gammazero: Probably, but I did not know how environ vars work with that.  Will an environ var like "tag1, tag2" be converted to ["tag1", "tag2"] ?

willscott: i think you want https://github.com/urfave/cli/blob/master/flag_string_slice.go#L72 "StringSliceFlag"
and yes, it should auto do the splitting of commas then

gammazero: Yes, just found out that was there.  Thank you.

controller/state/statedb.go
```diff
@@ -277,20 +279,42 @@ func (s *stateDB) AssignTask(ctx context.Context, req tasks.PopTask) (tasks.Task
 		return nil, fmt.Errorf("cannot assign %q status to task", req.Status.String())
 	}
 
+	var tags []interface{}
+	if req.Tags.Exists() {
+		for _, t := range strings.Split(req.Tags.Must().String(), ",") {
```
mvdan, line 284: why not make this a list of strings in IPLD directly?

tasks/gen.go
```diff
@@ -138,6 +140,7 @@ func main() {
 	ts.Accumulate(schema.SpawnStruct("PopTask", []schema.StructField{
 		schema.SpawnStructField("Status", "Status", false, false),
 		schema.SpawnStructField("WorkedBy", "String", false, false),
+		schema.SpawnStructField("Tags", "String", true, false),
```
mvdan, line 143: I think this should be called AnyTags or AllTags, to clarify if this is "at least one tag is shared with the task to pop" or "all tags must be shared with the task to pop".

mvdan (approved):
SGTM. I think you should make the matching semantics clearer though, because it's not terribly obvious to me from reading the code.

I think we probably want "all tags in a PopTask must be contained in the Tags list of the task being popped", i.e. also allowing extra tags to be present in the task.

tasks/gen.go
mvdan: Oh, and assuming this is a list too, I'd also make it an actual IPLD list.

gammazero: Done

controller/state/statedb.go
gammazero: Done

gammazero:
@mvdan Added comments to describe tag semantics.

willscott (approved):

PR #137, opened 2021-05-21
Add miner flag to queue-storage

hannahhoward:
queue-storage expects a miner but it's not currently in the flag list

note: discovered using dealbot -- not sure when this broke, fixing now


willscott (approved):

hannahhoward:
one sec on this -- i want to move our integration tests to use this rather than CURL

PR #140, opened 2021-05-24
fix(retrieval_task): properly consume ClientRetrieveWithEvents

hannahhoward:
# Goals

I noticed the retrievals were not recording the error messages and two deals were marked successful while seemingly never having been run. Investigation led me to understand ClientRetrieveWithEvents from the Lotus API behaves quite differently than one might expect always sending one last status message with the final error, and not sending any error status until then.

# Implementation

Essentially I modelled this after the existing LOTUS cli command that consumes ClientRetrieveWithEvents -- https://github.com/filecoin-project/lotus/blob/master/cli/client.go#L1187

Also: I am not sure the best way to test issues like this, other than by hand on nerpa or mainnet -- it's extremely difficult to actually simulate the errors that triggered this out of production.

### DIFF
--- a/tasks/retrieval_deal.go
+++ b/tasks/retrieval_deal.go
@@ -39,7 +39,7 @@ func MakeRetrievalDeal(ctx context.Context, config NodeConfig, node api.FullNode
 	if err != nil {
 		return err
 	}
-	return de.executeAndMonitorDeal(updateStage)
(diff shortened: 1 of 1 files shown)


hannahhoward:
Verified on US dealbot

willscott (approved):

PR #144, opened 2021-05-25
add graphiql index to graphql endpoint

### LINKED ISSUE #143: add graphql web console (graphiql)
currently hard to know how to form queries

willscott:
fix #143

### DIFF
--- a/controller/graphql/server.go
+++ b/controller/graphql/server.go
@@ -1,6 +1,7 @@
 package graphql
 
 import (
+	"embed"
 	"encoding/json"
 	"log"
 	"net/http"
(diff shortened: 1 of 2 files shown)


controller/graphql/server.go
```diff
@@ -10,6 +11,9 @@ import (
 	"github.com/graphql-go/graphql"
 )
 
+//go:embed index.html
+var index embed.FS
```
mvdan, line 15: I guess it's easier to use a FS for this, even if it's just one file, so we can use http.FileServer?

mvdan (approved):
I was about to say that graphiql was a typo for graphql :P

controller/graphql/server.go
willscott: right. otherwise it's a custom handler wrapping sending the bytes, right?

hannahhoward (approved):
LGTM

Issue #134, opened 2021-05-21
Status-specific Timeouts


mgoelzer:
This is part of the timeouts plan here:  https://hackmd.io/w4qrkgdDR56joFisnn85pA?view#Status-specific-Timeouts

It hasn't been implemented yet.

I think we could probably set a 30- or 60-min max on StatusOngoing during a retrieval.  I don't think a round trip of the `payment -> transfer -> new payment` loop is every more than 1 Gb, but would need to double check.

willscott:
fixed in #147

PR #147, opened 2021-05-25
Stage timeouts

gammazero:
When running the retrieval-deal or daemon command, specify deal stage timeouts using the `--stage-timeout` flag.  This flag can be specified multiple times to configure timeouts for deal stages.  For example:
```sh
./dealbot daemon --stage-timeout ProposeRetrieval=30m --stage-timeout DealAccepted=2h ...
```

The stage names are case-insensitive.  The timeout duration must be an integer with a suffix of "s", "m", or "h" to specify the time units.

This addresses issue #134

### DIFF
--- a/commands/flags.go
+++ b/commands/flags.go
@@ -53,6 +53,11 @@ var RetrievalFlags = []cli.Flag{
 		Usage: "payload cid to fetch from miner",
 		Value: "bafykbzacedikkmeotawrxqquthryw3cijaonobygdp7fb5bujhuos6wdkwomm",
 	},
+	altsrc.NewStringSliceFlag(&cli.StringSliceFlag{
+		Name:    "stage-timeout",
+		Usage:   "stagename=duration exampple: DealAccepted=30m",
+		EnvVars: []string{"STAGE_TIMEOUT"},
+	}),
 }
 
 var StorageFlags = []cli.Flag{
@@ -89,6 +94,11 @@ var StorageFlags = []cli.Flag{
 		EnvVars: []string{"DEALBOT_START_OFFSET"},
 		Value:   30760,
 	},
+	altsrc.NewStringSliceFlag(&cli.StringSliceFlag{
+		Name:    "stage-timeout",
+		Usage:   "stagename=duration exampple: DealAccepted=30m",
+		EnvVars: []string{"STAGE_TIMEOUT"},
+	}),
 }
 
 var SingleTaskFlags = append(DealFlags, MinerFlags...)
@@ -125,6 +135,11 @@ var DaemonFlags = append(DealFlags, append(CommonFlags, append(EndpointFlags, []
 		Name:    "tags",
 		Usage:   "comma separated tag strings",
 		EnvVars: []string{"DEALBOT_TAGS"},
+	}),
+	altsrc.NewStringSliceFlag(&cli.StringSliceFlag{
+		Name:    "stage-timeout",
+		Usage:   "stagename=duration exampple: DealAccepted=30m",
+		EnvVars: []string{"STAGE_TIMEOUT"},
 	}),
 }...)...)...)
 
--- a/commands/flags.go
+++ b/commands/flags.go
@@ -55,7 +55,7 @@ var RetrievalFlags = []cli.Flag{
 	},
 	altsrc.NewStringSliceFlag(&cli.StringSliceFlag{
 		Name:    "stage-timeout",
-		Usage:   "stagename=duration (example: DealAccepted=15m), set default: default=30m",
+		Usage:   "stagename=duration (example: DealAccepted=15m), DefaultRetrieval sets default",
 		EnvVars: []string{"STAGE_TIMEOUT"},
 	}),
 }
@@ -94,6 +94,11 @@ var StorageFlags = []cli.Flag{
 		EnvVars: []string{"DEALBOT_START_OFFSET"},
 		Value:   30760,
 	},
+	altsrc.NewStringSliceFlag(&cli.StringSliceFlag{
+		Name:    "stage-timeout",
+		Usage:   "stagename=duration (example: ProposeDeal=15m), DefaultStorage sets default",
+		EnvVars: []string{"STAGE_TIMEOUT"},
+	}),
 }
 
 var SingleTaskFlags = append(DealFlags, MinerFlags...)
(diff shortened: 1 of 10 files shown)


commands/flags.go
```diff
@@ -53,6 +53,11 @@ var RetrievalFlags = []cli.Flag{
 		Usage: "payload cid to fetch from miner",
 		Value: "bafykbzacedikkmeotawrxqquthryw3cijaonobygdp7fb5bujhuos6wdkwomm",
 	},
+	altsrc.NewStringSliceFlag(&cli.StringSliceFlag{
+		Name:    "stage-timeout",
+		Usage:   "stagename=duration exampple: DealAccepted=30m",
```
rvagg, line 58:
```suggestion
		Usage:   "stagename=duration (example: DealAccepted=30m)",
```

commands/flags.go
```diff
@@ -126,6 +131,11 @@ var DaemonFlags = append(DealFlags, append(CommonFlags, append(EndpointFlags, []
 		Usage:   "comma separated tag strings",
 		EnvVars: []string{"DEALBOT_TAGS"},
 	}),
+	altsrc.NewStringSliceFlag(&cli.StringSliceFlag{
+		Name:    "stage-timeout",
+		Usage:   "stagename=duration exampple: DealAccepted=30m",
```
rvagg, line 136:
```suggestion
		Usage:   "stagename=duration (example: DealAccepted=30m)",
```

tasks/retrieval_deal.go
```diff
@@ -174,3 +198,43 @@ func (rp *_RetrievalTask__Prototype) Of(minerParam string, payloadCid string, ca
...
+		if len(parts) != 2 {
+			return nil, fmt.Errorf("invalid stage timeout specification: %s", spec)
+		}
+		stage := parts[0]
+		timeout := parts[1]
+		d, err := time.ParseDuration(strings.TrimSpace(timeout))
+		if err != nil || d < time.Second {
+			return nil, fmt.Errorf("invalid value for stage %q timeout: %s", stage, timeout)
+		}
+		timeouts[strings.ToLower(strings.TrimSpace(stage))] = d
```
mvdan, line 220: you probably want to error if there are duplicates

tasks/retrieval_deal.go
```diff
@@ -174,3 +198,43 @@ func (rp *_RetrievalTask__Prototype) Of(minerParam string, payloadCid string, ca
...
+	for stageName, _ := range RetrievalStages {
+		stageName = strings.ToLower(stageName)
+		d, ok := timeouts[stageName]
+		if ok {
+			if stageTimeouts == nil {
+				stageTimeouts = make(map[string]time.Duration)
+			}
+			stageTimeouts[stageName] = d
+		}
+	}
```
mvdan, line 233: hmm, so if I make a typo in specifying a stage name, it won't be seen by the iteration over RetrievalStages and it will be ignored?

I think you should error if any input stage timeouts end up being unused because we don't know any stage by that name.

tasks/retrieval_deal.go
```diff
@@ -85,15 +89,32 @@ func (de *retrievalDealExecutor) executeAndMonitorDeal(ctx context.Context, upda
...
+	err = updateStage(stage, dealStage)
 	if err != nil {
 		return err
 	}
 
 	lastStatus := retrievalmarket.DealStatusNew
-	var lastBytesReceived uint64 = 0
+	var lastBytesReceived uint64
+	var prevStage string
+	var stageTimeout <-chan time.Time
```
mvdan, line 100: the chan can be declared inside the loop, no? you can then avoid the `stageTimeout = nil` assignment too, as it's already the zero value.

mvdan (approved):

tasks/retrieval_deal.go
willscott: not all deal flows will go through all stages / paths i think, which may complicate that

tasks/retrieval_deal.go
```diff
@@ -85,15 +89,32 @@ func (de *retrievalDealExecutor) executeAndMonitorDeal(ctx context.Context, upda
...
+
 	for {
+		if stage != prevStage {
+			// If there is a timeout for the current deal stage, then set the timer
+			if timeout, ok := stageTimeouts[strings.ToLower(stage)]; ok {
+				stageTimeout = time.After(timeout)
+			} else {
+				stageTimeout = nil
```
willscott, line 108: can we add a flag for a default stage timeout so we don't get stuck in an unknown stage forever?

tasks/retrieval_deal.go
mvdan: this would also make the code easier to follow - the chan var being shared by different iterations initially made me think that the channel would be reused, when I'm pretty sure it's not

willscott:
is it not possible to time out on storage deals? what about when timing out on trying to dial the miner to connect to it?

tasks/retrieval_deal.go
gammazero: This is done by setting `--stage-timeout default:30m`.  Noted in flag usage.

tasks/retrieval_deal.go
gammazero: Done

tasks/retrieval_deal.go
gammazero: Will return error if a stage timeout is unusable.  It is still possible that some deals may not use some stage timeouts, but this will at least report any user input that cannot be used.

Done

tasks/retrieval_deal.go
gammazero: Done

commands/flags.go
gammazero: Done

commands/flags.go
gammazero: Done

gammazero:
> is it not possible to time out on storage deals? what about when timing out on trying to dial the miner to connect to it?

Currently, only the overall timeout (24h) applies to storage deals.  I can easily add storage deal timeouts, or at least add them for [common stages](https://github.com/filecoin-project/dealbot/blob/main/tasks/model.go#L94-L101).

While retrieval deals have a default stage timeout of 30m,  should this same or any default apply to storage deals?

mvdan (approved):
Maybe we want to document how we arrived at the 30m/3h default timeouts? Otherwise LGTM!

PR #153, opened 2021-05-26
test(state): add a test demonstrating patch/get parallelism issue

hannahhoward:
# Goals

Demonstrate failure that occurs from mix of Update & GetAll

# Implementation

The added test reliably creates a "transaction within transacction error" on my machine.

Note that if you comment out either operation (and reduce total expected error channel reads to 50), it passes, so it is the combination of Update + GetAll that triggers this.

### DIFF
--- a/controller/state/statedb_test.go
+++ b/controller/state/statedb_test.go
@@ -414,6 +414,68 @@ func TestUpdateTasks(t *testing.T) {
 	}
 }
 
+func TestUpdateAndQueryTasks(t *testing.T) {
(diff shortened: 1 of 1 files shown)


gammazero:
Rebase, and I bet it will start working.

hannahhoward:
@willscott should we merge this or call it a useful thing that helped us debug and let it go?

willscott:
i'd guess we're not going to catch any further issues with this test - it's more likely to be a new / different method that doesn't properly wrap itself as a txn? so i guess i'd let it go.

PR #155, opened 2021-05-26
Handle all transactions using the transaction wrapper

gammazero:
- Fix intermittent test fail, due to varying ordering of map
- Wrap additional functions in transaction

### DIFF
--- a/controller/state/statedb.go
+++ b/controller/state/statedb.go
@@ -177,17 +177,19 @@ type sdbstore struct {
 }
 
 func (s *sdbstore) Get(c cid.Cid) (blockformat.Block, error) {
-	tx, err := s.stateDB.db().Begin()
-	if err != nil {
-		return nil, err
-	}
-	defer tx.Commit()
-	loader := txContextLoader(s.Context, tx)
-	blkReader, err := loader(cidlink.Link{Cid: c}, ipld.LinkContext{})
-	if err != nil {
-		return nil, err
-	}
-	data, err := ioutil.ReadAll(blkReader)
+	var data []byte
+	err := s.stateDB.transact(s.Context, func(tx *sql.Tx) error {
+		loader := txContextLoader(s.Context, tx)
+		blkReader, err := loader(cidlink.Link{Cid: c}, ipld.LinkContext{})
+		if err != nil {
+			return err
+		}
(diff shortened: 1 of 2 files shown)


controller/state/statedb.go
```diff
@@ -632,27 +647,25 @@ func (s *stateDB) countTasks(ctx context.Context) (int, error) {
...
+		cidLink, err := cid.Decode(c)
+		if err != nil {
+			return err
+		}
 
-	if err := tx.Commit(); err != nil {
+		loader := txContextLoader(ctx, tx)
+		if err = (cidlink.Link{Cid: cidLink}).Load(ctx, ipld.LinkContext{}, na, loader); err != nil {
+			return err
+		}
```
willscott, line 665: this can lead to a sub-txn, and should be done outside of this txn, i think?

hannahhoward: I don't think so no? cause the txContextLoader takes the passed in transaction.

controller/state/statedb.go
```diff
@@ -511,26 +519,33 @@ func (s *stateDB) NewRetrievalTask(ctx context.Context, retrievalTask tasks.Retr
...
-		if err = rows.Scan(&status, &stage, &run, &ts); err != nil {
-			return nil, err
+	err := s.transact(ctx, func(tx *sql.Tx) error {
+		rows, err := tx.QueryContext(ctx, taskHistorySQL, taskID)
+		if err != nil {
+			return err
+		}
+		defer rows.Close()
+
+		history = nil
```
hannahhoward, line 530: Non-blocking:
I assume this line is cause `transact` may retry -- we should probably indicate somewhere that retries are possible and if you're using closures to access higher scoped vars you better reset them. took me a few rereads to figure out what this mean or why it was needed.

hannahhoward (approved):
LGTM with a couple comments

controller/state/statedb.go
gammazero: Nice deduction!  That's exactly why.  I will add comments.

gammazero: Done

willscott (approved):

hannahhoward (approved):
LGTM, part 2, let's merge it!

PR #166, opened 2021-05-29
deploy mainnet on tags

coryschwartz:
Continue deploying to nerpanet for every merge to the main branch.
Deploy to mainnet when there are git tags versioned releases

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -60,6 +60,12 @@ jobs:
         type: string
       release:
         type: string
+      circle_context:
+        type: string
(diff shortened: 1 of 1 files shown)


coryschwartz:
the "dev" cluster is the default environment where things are deployed unless otherwise specified so the other parameters weren't needed before. Now we're deploying under different conditions to two different kubernetes clusters on two AWS accounts. 

I'll create another PR to add a github action, which is completely optional, but can be used for triggering deploys manually wihtout the need to merge.

willscott (approved):

PR #167, opened 2021-05-29
github action

coryschwartz:
This will give you a nice-ish-looking UI that can be used to trigger deploys. You'll find it in the "actions" tab

### DIFF
--- a/.github/workflows/trigger-helm-deploy.yaml
+++ b/.github/workflows/trigger-helm-deploy.yaml
@@ -1,0 +1,64 @@
+---
+name: trigger-helm-deploy
+on:
+  workflow_dispatch:
+    inputs:
+      circle_context:
+        description: "CI context for aws credentials"
+        default: "sentinel-staging-deploy"
+      kubernetes_cluster:
+        description: "kubernetes cluster"
+        default: "mainnet-us-east-2-dev-eks"
+      aws_region:
+        description: "aws region"
+        default: "us-east-2"
+      namespace:
+        description: "kubernetes namespace"
+        default: "ntwk-nerpanet-dealbot"
+      release:
+        description: "helm release"
+        default: "dealbot-0"
+      chart:
+        description: "the helm chart"
+        default: "filecoin/dealbot"
+      chart_version:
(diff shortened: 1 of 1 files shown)


.github/workflows/trigger-helm-deploy.yaml
```diff
@@ -0,0 +1,64 @@
...
+      circle_branch:
+        description: "circleci branch (used for testing new lotus-infra deploy code)"
+        default: master
+
+jobs:
+  trigger-circleci:
+    runs-on: ubuntu-latest
+    steps:
+      - name: trigger
+        uses: promiseofcake/circleci-trigger-action@v1
```
willscott, line 45: should we pin this to a specific commit?

willscott (approved):

.github/workflows/trigger-helm-deploy.yaml
gammazero: The version tag should be sufficient.

willscott: see e.g. https://github.com/ipld/.github/pull/18 - i believe we tried to have a policy of pinning 3rd party actions by hash

mvdan: Right, you don't want to give that user a way to run arbitrary code in our CI, especially if there's secrets involved :)

.github/workflows/trigger-helm-deploy.yaml
```diff
@@ -0,0 +1,64 @@
...
+      override_repository:
+        description: "helm chart image.repository"
+      override_tag:
+        description: "helm chart image.tag"
+      slack_channel:
+        description: "the slack channel where notifications are sent"
+        default: "netops-circleci-test"
+      circle_branch:
+        description: "circleci branch (used for testing new lotus-infra deploy code)"
+        default: master
```
willscott, line 38:
```suggestion
        default: main
```

gammazero, line 38: SHould this be `main`

Issue #160, opened 2021-05-28
a dealbot daemon with a single worker polls for more work while doing a deal


mvdan:
If you run the second integration test via `bash integration_tests/02_controller_daemon.sh`, while the deal is being made (which takes multiple minutes), the controller gets spammed with pop-task requests every five seconds:

```
dealbot-daemon.log: 2021-05-28T15:48:30.529+0100        INFO    engine  tasks/storage_deal.go:298       Deal status     {"cid": "bafyreibekranctfmu4r4f3djo4tix4uhcusqn3ptejs47qoverehc56fsi", "piece": "baga6ea4seaqnvupw3dnivjfy5wdz4wfr7u35lkqkrwtb22joj2xwz6nnjuxocki", "state": "StorageDealAwaitingPreCommit", "message": "", "provider": "f01000"}
dealbot-controller.log: 127.0.0.1 - - [28/May/2021:15:48:35 +0100] "POST /pop-task HTTP/1.1" 204 0
dealbot-controller.log: 127.0.0.1 - - [28/May/2021:15:48:40 +0100] "POST /pop-task HTTP/1.1" 204 0
dealbot-controller.log: 127.0.0.1 - - [28/May/2021:15:48:45 +0100] "POST /pop-task HTTP/1.1" 204 0
dealbot-controller.log: 127.0.0.1 - - [28/May/2021:15:48:50 +0100] "POST /pop-task HTTP/1.1" 204 0
dealbot-controller.log: 127.0.0.1 - - [28/May/2021:15:48:55 +0100] "POST /pop-task HTTP/1.1" 204 0
dealbot-controller.log: 127.0.0.1 - - [28/May/2021:15:49:00 +0100] "POST /pop-task HTTP/1.1" 204 0
dealbot-daemon.log: 2021-05-28T15:49:02.023+0100        INFO    engine  tasks/storage_deal.go:298       Deal status     {"cid": "bafyreibekranctfmu4r4f3djo4tix4uhcusqn3ptejs47qoverehc56fsi", "piece": "baga6ea4seaqnvupw3dnivjfy5wdz4wfr7u35lkqkrwtb22joj2xwz6nnjuxocki", "state": "StorageDealSealing", "message": "", "provider": "f01000"}
```

This feels wrong. The second integration test runs with a single worker goroutine, and it's clearly busy doing a task, so it should not be attempting to pick up a second one.

willscott:
[...]

PR #170, opened 2021-06-02
Scheduler update

gammazero:
- Fix job ID race.
- No additional goroutine for scheduled jobs
- Do not pop task unless worker is available

Addresses issue #160

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -35,6 +35,7 @@ type Engine struct {
 	shutdown   chan struct{}
 	stopped    chan struct{}
 	tags       []string
+	workerPing chan struct{}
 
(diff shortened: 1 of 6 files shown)


willscott (approved):
would be great to add a test for this in one of the integration tests:
* 2 job available
* 1 worker starts up and claims 1
* wait a bit
* make sure the other one is still available and hasn't accidentally been claimed

gammazero:
@willscott I added functionality to the 02 integration test to check that popTask is not called while the worker is busy.  It accomplishes the intent of the test you suggested, but works whether or not additional tasks are present.

willscott (approved):

PR #180, opened 2021-06-04
Minfil

willscott:
flags on daemon to not accept tasks when not running against a lotus instance that can run them

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -327,5 +327,20 @@ func (e *Engine) apiGood(ctx context.Context) bool {
 	if localBalance.LessThan(e.nodeConfig.MinWalletBalance) {
 		return false
 	}
+
+	head, err := e.node.ChainHead(ctx)
+	if err != nil {
+		log.Errorw("could not query api for head", "error", err)
+		return false
+	}
+	localCap, err := e.node.StateVerifiedClientStatus(ctx, e.nodeConfig.WalletAddress, head.Key())
+	if err != nil {
+		log.Errorw("could not query api for datacap", "error", err)
+		return false
+	}
+	if localCap.LessThan(e.nodeConfig.MinWalletCap) {
+		return false
+	}
+
 	return true
 }
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -107,7 +107,8 @@ taskLoop:
 		default:
 		}
 
-		if e.pingWorker() {
+		check, _ := context.WithDeadline(context.Background(), time.Now().Add(3*time.Second))
+		if e.pingWorker() && e.apiGood(check) {
 			// Check if there is a new task
 			task := e.popTask()
 			if task != nil {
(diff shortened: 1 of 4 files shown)


lotus/client.go
```diff
@@ -58,11 +59,27 @@ func SetupClientFromCLI(cctx *cli.Context) (tasks.NodeConfig, api.FullNode, Node
 	if err != nil {
 		return tasks.NodeConfig{}, nil, nil, fmt.Errorf("wallet is not a Filecoin address: %s, %s", cctx.String("wallet"), err)
 	}
+	mf := big.NewInt(100000000000000)
```
mvdan, line 62: perhaps document what this number is?

lotus/client.go
```diff
@@ -108,10 +125,26 @@ func SetupClient(ctx context.Context, cliCtx *cli.Context) (tasks.NodeConfig, ap
 	if err != nil {
 		return tasks.NodeConfig{}, nil, nil, fmt.Errorf("wallet is not a Filecoin address: %s, %s", wallet, err)
 	}
+	mf := big.NewInt(100000000000000)
```
mvdan, line 128: ditto

engine/engine.go
```diff
@@ -107,7 +107,8 @@ taskLoop:
 		default:
 		}
 
-		if e.pingWorker() {
+		check, _ := context.WithDeadline(context.Background(), time.Now().Add(3*time.Second))
```
mvdan, line 110: easier with WithTimeout, since you avoid the time.Now().Add bit

mvdan (approved):

engine/engine.go
gammazero, line 110: Since this context is only used within `apiGood()`, recommend creating it in there.  Also, do not ignore the cancel function so that cancel can be called as soon as ctx is not needed and its goroutine will be ended.

engine/engine.go
```diff
@@ -315,3 +316,34 @@ func (e *Engine) pingWorker() bool {
...
+
+// apiGood returns true if the api can be reached and reports sufficient fil/cap to process tasks.
+func (e *Engine) apiGood(ctx context.Context) bool {
+	localBalance, err := e.node.WalletBalance(ctx, e.nodeConfig.WalletAddress)
+	if err != nil {
+		log.Errorw("could not query api for wallet balance", "error", err)
+		return false
+	}
+	if localBalance.LessThan(e.nodeConfig.MinWalletBalance) {
+		return false
```
gammazero, line 328: Even though this is not an error, perhaps logging a warning would be useful to alert admins that they need to rotate/fund the wallet.

lotus/client.go
gammazero, line 62: This number `100000000000000` is used in multiple places. Suggest using a `const`.

gammazero (commented):
Some minor change suggestions

PR #188, opened 2021-06-08
update graphql to have more useful error messages

willscott:
schema is an auto-generated file

### DIFF
--- a/controller/graphql/schema.go
+++ b/controller/graphql/schema.go
@@ -19,6 +19,7 @@ const nodeLoaderCtxKey = "NodeLoader"
 var errNotNode = fmt.Errorf("Not IPLD Node")
 var errInvalidLoader = fmt.Errorf("Invalid Loader Provided")
 var errInvalidLink = fmt.Errorf("Invalid link")
+var errUnexpectedType = "Unexpected type %T. expected %s"
 
 func resolve_map_at(p graphql.ResolveParams) (interface{}, error) {
     ts, ok := p.Source.(ipld.Node)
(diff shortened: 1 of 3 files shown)


tasks/gen.go
```diff
@@ -90,10 +90,10 @@ func main() {
...
 		schema.SpawnStructField("PayloadCID", "String", false, false),
 		schema.SpawnStructField("CARExport", "Bool", false, false),
-		schema.SpawnStructField("Schedule", "String", true, false),
-		schema.SpawnStructField("ScheduleLimit", "String", true, false),
-		schema.SpawnStructField("Tag", "String", true, false),
-		schema.SpawnStructField("MaxPriceAttoFIL", "Int", true, false),
+		schema.SpawnStructField("Schedule", "String", true, true),
+		schema.SpawnStructField("ScheduleLimit", "String", true, true),
+		schema.SpawnStructField("Tag", "String", true, true),
+		schema.SpawnStructField("MaxPriceAttoFIL", "Int", true, true),
```
mvdan, line 96: I assume we're all in agreement this is a backwards compatible change in terms of encoded data, since we simply add an extra option for these fields to be `null`. Just thinking about it outloud.

controller/graphql/server.go
```diff
@@ -94,11 +131,26 @@ func GetHandler(db state.State, accessToken string) (*http.ServeMux, error) {
...
+		n := builder.Build()
+		return n, nil
+	}
+
 	mux := http.NewServeMux()
 	mux.Handle("/", http.FileServer(http.FS(index)))
 	mux.Handle("/graphql", CorsMiddleware(func(w http.ResponseWriter, r *http.Request) {
 		var result *graphql.Result
-		ctx := r.Context()
+		sctx := r.Context()
```
mvdan, line 152: I guess you don't need this variable? having fewer names in scope makes it easier for me to read the code at least, since I need to worry about fewer names at any point.

mvdan (approved):

tasks/gen.go
mvdan: Ah I see. I replied to your ipld-prime thread. We already have live data, so I agree with your fix here - the alternative would have been to use Representation consistently with codecs.

PR #190, opened 2021-06-08
add a posthook that can be specified

### LINKED ISSUE #172: Pre-finalization script hook
We have a use case for both Slingshot and Discover where we need to run a script after a retrieval is completed but before the task is finalized.  (The purpose of this is to do an `ipfs pin add` command on the newly retrieved script.)  The script would be run by an engine instance:  

 - on stdin the script would get basic parameters like the CID and miner ID
 - the script would return an exit code of 0 (success) or non-zero for error
 - for non-zero return values, the task would be finalized as a client error with the last line of the script's stdout going into `ErrorMessage` field on the finalized task

Maybe there is a better way to generalize this so it can apply to many different use cases, including ones we have no thought of yet.

willscott:
fix #172

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -2,6 +2,7 @@ package engine
 
 import (
 	"context"
+	"os/exec"
 	"sync"
 	"time"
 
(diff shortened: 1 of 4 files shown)


engine/engine.go
```diff
@@ -268,6 +269,12 @@ func (e *Engine) runTask(ctx context.Context, task tasks.Task, runCount, worker
 		}
 		tlog.Errorw("error updating final status", "err", err)
 	}
+	if e.nodeConfig.PostHook != "" {
+		err := exec.Command("/bin/bash", e.nodeConfig.PostHook, task.GetUUID()).Run()
```
mvdan, line 273: you mean PostHook to be a script file, not a command, right? if so, I'd document the flag as a script, not a command. otherwise I'd expect to give it `--posthook="some command"`, not `--posthook=/path/to/some-command.sh`.

PR #191, opened 2021-06-08
manually attempt connection / simulate a rtt

willscott:
First pass at a latency probe

* we should see if this attempt at reading (the only thing that actually blocks down low enough to need a libp2p rtt) and then closing is a reasonable way to pull out an RTT latency
* we should make sure this can't error in unexpected ways.

### DIFF
--- a/netutil/netutil.go
+++ b/netutil/netutil.go
@@ -1,0 +1,56 @@
+package netutil
+
+import (
+	"io"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/libp2p/go-libp2p"
+	"github.com/libp2p/go-libp2p-core/peer"
+	"github.com/multiformats/go-multiaddr"
+	"golang.org/x/net/context"
+)
+
+func TryAcquireLatency(ctx context.Context, ai peer.AddrInfo) (multiaddr.Multiaddr, int64, error) {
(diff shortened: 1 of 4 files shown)


netutil/netutil.go
```diff
@@ -0,0 +1,53 @@
...
+			end = time.Now()
+		} else {
+			err = serr
+		}
+	}()
+
+	// send a write of random bytes, which should make the other side unhappy.
+	stream.Write([]byte("-latencytest-\n\x00"))
+	wg.Wait()
+	defer conn.Close()
```
gammazero, line 50: Does not seem to be a need to defer here.

netutil/netutil.go
```diff
@@ -0,0 +1,53 @@
...
+
```
gammazero, line 51:
```go
if err != nil {
   return nil, 0, err
}
```

gammazero (changes requested):

gammazero (approved):

netutil/netutil.go
willscott: we may still get a useful remote node multiaddr by this point, which is why i'm tempted to still return it even if the error indicates the measured latency shouldn't be trusted

hannahhoward (approved):
LGTM

PR #194, opened 2021-06-09
tasks: record first and last byte for retrievals

mvdan:
(see commit message)


mvdan:
Probably a good idea to read the commit message before the diff.

willscott (approved):

PR #196, opened 2021-06-09
Fix Netdiag function

hannahhoward:
# Goals

Fix an issue that cause MinerLatency not to show up. 

# Implementation

The miner latency was only getting recorded when there was an error.

Also, this ticket began as an effort to verify the logs generated by actually unit testing some of the code that talks to Lotus. As you can see, while not totally ideal, it's definitely possible to use the generated Lotus Mock API (now part of the Lotus codebase and always up to date) to write unit tests that "talk" to Lotus.

### DIFF
--- a/tasks/storage_deal.go
+++ b/tasks/storage_deal.go
@@ -21,7 +21,10 @@ import (
 	"github.com/filecoin-project/lotus/chain/types"
 	"github.com/google/uuid"
 	"github.com/ipld/go-ipld-prime/schema"
+	"github.com/libp2p/go-libp2p"
+	"github.com/libp2p/go-libp2p-core/host"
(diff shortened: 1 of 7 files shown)


tasks/storage_deal.go
```diff
@@ -170,8 +175,8 @@ func (de *dealExecutor) netDiag(l logFunc) error {
 		return err
 	}
 
-	peerAddr, peerL, err := netutil.TryAcquireLatency(de.ctx, de.pi)
-	if err != nil {
+	peerAddr, peerL, err := netutil.TryAcquireLatency(de.ctx, de.pi, de.makeHost)
+	if err == nil {
```
willscott, line 179: 🙃

willscott (approved):
thanks for catching that!

PR #200, opened 2021-06-10
Run tests in parallel

hannahhoward:
# Goals

Faster builds

# Implementation

- split tests via circle so they run in parallel

# For Discussion

Pros:
- shaves 10 min + off the total time through a build-push
Cons:
- adds a lot of complexity to the CI configuration

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,34 +9,102 @@ executors:
     resource_class: 2xlarge
 
 commands:
+  save-cache:
+    description: Save the go mod & build cache.
+    steps:
+      - save_cache:
+          key: >-
+            v3-gosrc-{{ .Branch }}-{{ checksum "go.sum" }}-{{
+            .Environment.CIRCLE_JOB }}
+          name: Saving Source Cache
+          paths:
+            - /home/circleci/go/pkg/mod
+      - save_cache:
+          key: >-
+            v3-gobin-{{ arch }}-{{ .Branch }}-{{ checksum "go.sum" }}-{{
+            .Environment.CIRCLE_JOB }}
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -9,34 +9,102 @@ executors:
...
       - run: sudo apt-get install ocl-icd-opencl-dev libhwloc-dev
+  build-lotus:
+    steps:
       - run: git clone --depth=1 https://github.com/filecoin-project/lotus /tmp/lotus
       - run: cd /tmp/lotus && git submodule update --init --depth=1
-      - run: cd /tmp/lotus && make debug && sudo make install
+      - run: cd /tmp/lotus && make debug
+  build-dealbot:
+    steps:
+      - run: go install ./...
```
willscott, line 64: is there not a pre-built lotus docker image we could use as our base rather than doing the lotus compile in each CI?

hannahhoward: you would think so... but you might not be correct :(

hannahhoward: One would think so... but... not super obvious.

Also we need a couple of the obscure executables and a debug build... that's not the standard. I think we could build our own, or use caching. One question is whether we want to stay up to date with lotus master or accept a fixed version that could run out of date.

willscott: caching would be reasonable as a stop-gap. I'd like to not have CI wait to re-build all of lotus every commit. that seems like a recipe for disaster where we keep fixing the lotus bugs.

willscott (approved):

willscott:
we disabled the long test, right @hannahhoward ? does it still make sense to merge this?

PR #202, opened 2021-06-10
Skip 03_scheduled_task integration test

gammazero:
Skip the scheduled task integration test for now.  Other tests check that dealbot is able to make deals, so the only added value of this test is to check that a scheduled task is re-run due to having a schedule.


willscott (commented):
can we disable at the circleci config layer, rather than at the bash level?

willscott:
yeah. the balance though is that it's harder to run manually

PR #203, opened 2021-06-10
multiretrieve

willscott:
saves files / folders retrieved by their piece cid so multiple can happen at once


hannahhoward (approved):
LGTM

PR #204, opened 2021-06-10
Allow repeated retrieval of the same data

gammazero:
Delete the local file if it already exists to fore re-retrieval.

Draft, because I have not confirmed if this is doing what is needed, other than seeing that the miner log is slightly longer.
@hannahhoward - Is the presence of this file what lotus uses to determine whether or not to fetch from miner?


gammazero:
Does not appear to be necessary

PR #220, opened 2021-06-15
allow for `dealbot drain` cli that exits once a host is fully done

willscott:
this seems to be the component that works well with something like https://blog.gruntwork.io/gracefully-shutting-down-pods-in-a-kubernetes-cluster-328aecec90d

### DIFF
--- a/commands/drain.go
+++ b/commands/drain.go
@@ -1,0 +1,53 @@
+package commands
+
+import (
+	"context"
+	"time"
+
+	"github.com/filecoin-project/dealbot/controller/client"
+	"github.com/urfave/cli/v2"
+)
+
+var DrainCmd = &cli.Command{
+	Name:   "drain",
+	Usage:  "Stop a dealbot daemon",
+	Flags:  append(EndpointFlags, idFlag),
+	Action: drainCommand,
+}
+
+func drainCommand(c *cli.Context) error {
+	host_id := c.String("id")
+	controller := client.New(c)
+
+	// drain `host_id`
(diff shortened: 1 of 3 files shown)


commands/drain.go
```diff
@@ -0,0 +1,53 @@
...
+	controller := client.New(c)
+
+	// drain `host_id`
+	if err := controller.Drain(c.Context, host_id); err != nil {
+		return err
+	}
+
+	// poll until tasks from `host_id` are done.
+	for {
+		if isActive(c.Context, controller, host_id) {
```
mvdan, line 29: `for isActive(...) { time.Sleep(...) }`?

commands/drain.go
```diff
@@ -0,0 +1,53 @@
...
+	"context"
+	"time"
+
+	"github.com/filecoin-project/dealbot/controller/client"
+	"github.com/urfave/cli/v2"
+)
+
+var DrainCmd = &cli.Command{
+	Name:   "drain",
+	Usage:  "Stop a dealbot daemon",
```
mvdan, line 13: I think you could make the usage text a bit clearer about this command waiting for tasks to finish, otherwise I'd assume the command would exit ~immediately, unless I was already familiar with the concept of "draining" :)

mvdan (approved):

hannahhoward (approved):
LGTM

hannahhoward:
@mvdan merging for silly reasons (want to get a new release tagged and this seems mostly reasonable) but we should go back and maybe address your comments later.

PR #224, opened 2021-06-16
hook for scheduling retrievals after successful stores

willscott:
* allow the `/tasks/<uuid>` endpoint to take a `?parsed` parameter to show derived fields instead of full logs
* add a bash script to filter to storage deals that have reached 'time to last byte' and then take the payload and request a retrieval from the same miner for that payload in a week.

### DIFF
--- a/controller/tasks.go
+++ b/controller/tasks.go
@@ -1,15 +1,18 @@
 package controller
 
 import (
+	"bytes"
 	"database/sql"
 	"errors"
 	"fmt"
+	"io"
 	"net/http"
 	"strconv"
 	"time"
 
 	"github.com/gorilla/mux"
 	"github.com/ipld/go-car"
+	"github.com/ipld/go-ipld-prime"
 	"github.com/ipld/go-ipld-prime/codec/dagjson"
 	basicnode "github.com/ipld/go-ipld-prime/node/basic"
 	"github.com/ipld/go-ipld-prime/traversal/selector"
@@ -275,8 +278,23 @@ func (c *Controller) getTaskHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	w.WriteHeader(http.StatusOK)
-	dagjson.Encoder(task.Representation(), w)
+	if _, set := vars["parsed"]; set {
+		nilStore := func(_ ipld.LinkContext) (io.Writer, ipld.StoreCommitter, error) {
+			b := bytes.Buffer{}
+			return &b, func(_ ipld.Link) error { return nil }, nil
+		}
+		finished, err := task.Finalize(r.Context(), nilStore)
+		if err != nil {
+			log.Errorw("finalize task error", "err", err.Error())
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+		w.WriteHeader(http.StatusOK)
+		dagjson.Encoder(finished, w)
+	} else {
+		w.WriteHeader(http.StatusOK)
+		dagjson.Encoder(task.Representation(), w)
+	}
 }
 
 func (c *Controller) carHandler(w http.ResponseWriter, r *http.Request) {
--- a/controller/tasks.go
+++ b/controller/tasks.go
@@ -1,7 +1,6 @@
 package controller
 
 import (
-	"bytes"
 	"database/sql"
 	"errors"
 	"fmt"
@@ -280,8 +279,7 @@ func (c *Controller) getTaskHandler(w http.ResponseWriter, r *http.Request) {
 
 	if _, set := vars["parsed"]; set {
 		nilStore := func(_ ipld.LinkContext) (io.Writer, ipld.StoreCommitter, error) {
-			b := bytes.Buffer{}
-			return &b, func(_ ipld.Link) error { return nil }, nil
+			return io.Discard, func(_ ipld.Link) error { return nil }, nil
 		}
 		finished, err := task.Finalize(r.Context(), nilStore)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


controller/tasks.go
```diff
@@ -275,8 +278,23 @@ func (c *Controller) getTaskHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	w.WriteHeader(http.StatusOK)
-	dagjson.Encoder(task.Representation(), w)
+	if _, set := vars["parsed"]; set {
+		nilStore := func(_ ipld.LinkContext) (io.Writer, ipld.StoreCommitter, error) {
+			b := bytes.Buffer{}
+			return &b, func(_ ipld.Link) error { return nil }, nil
```
mvdan, line 284: to use a "discard" writer, you can use io.Discard

controller/tasks.go
```diff
@@ -275,8 +278,23 @@ func (c *Controller) getTaskHandler(w http.ResponseWriter, r *http.Request) {
...
+		finished, err := task.Finalize(r.Context(), nilStore)
+		if err != nil {
+			log.Errorw("finalize task error", "err", err.Error())
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+		w.WriteHeader(http.StatusOK)
+		dagjson.Encoder(finished, w)
+	} else {
+		w.WriteHeader(http.StatusOK)
```
mvdan, line 295: I think we don't need this - I seem to recall that the net/http server already replies with a 200 if bytes are written to the response without an explicit response header.

scheduler/hooks/retrieve_after_store.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+# Usage: bash retrieve_after_store.sh <UUID>
+# expected installed env binaries: mktemp, curl, jq
+# env vars used:
+# DEALBOT_CONTROLLER_ENDPOINT
+
+curl -o /tmp/dealbot.$1 $DEALBOT_CONTROLLER_ENDPOINT/tasks/$1\?parsed
```
mvdan, line 7: pretty sure you don't need to escape the question mark FYI, it's not at all special in terms of shell syntax

scheduler/hooks/retrieve_after_store.sh
```diff
@@ -0,0 +1,24 @@
...
+
+if [ "$(cat /tmp/dealbot.$1 | jq '.StorageTask == null')" = "true" ]; then
```
mvdan, line 9: you can avoid the cat, like:

    if [ "$(jq '.StorageTask == null' /tmp/dealbot.$1)" = "true" ]; then

this is bash, so you could also use the safer and more intuitive `[[` builtin, which e.g. saves you from needing quotes for safety against word splitting:

    if [[ $(jq '.StorageTask == null' /tmp/dealbot.$1) == "true" ]]; then

scheduler/hooks/retrieve_after_store.sh
```diff
@@ -0,0 +1,24 @@
...
+fi
+
+if [ "$(cat /tmp/dealbot.$1 | jq '.TimeToLastByteMS == null')" = "true" ]; then
+  echo "task did not finish transfer."
+  exit 0
+fi
+
+MINER=$(cat /tmp/dealbot.$1 | jq .StorageTask.Miner -c)
+TAG=$(cat /tmp/dealbot.$1 | jq .StorageTask.Tag -c)
+PAYLOAD=$(cat /tmp/dealbot.$1 | jq .PayloadCID -c)
```
mvdan, line 21: same useless use of cat, if you're up for it :)

scheduler/hooks/retrieve_after_store.sh
```diff
@@ -0,0 +1,24 @@
...
+
+echo "{\"Miner\": $MINER, \"Tag\": $TAG, \"PayloadCID\": \"$PAYLOAD\", \"CarExport\": false, \"Schedule\": \"$(date -v+5M +"%M %H * * *")\", \"ScheduleLimit\": 1}" | jq -c . > /tmp/dealbot.$1.json
```
mvdan, line 23: hmm, not sure what `date` you used, but this seems to be non-standard and breaks on linux with GNU coreutils:

```
$ date -v+5M +"%M %H * * *"
date: invalid option -- 'v'
Try 'date --help' for more information.
```

scheduler/hooks/retrieve_after_store.sh
```diff
@@ -0,0 +1,24 @@
...
+curl -X POST -d "@/tmp/dealbot.$1.json" -H "Content-Type: application/json" $DEALBOT_CONTROLLER_ENDPOINT/tasks/retrieval
```
mvdan, line 24: I think this would be more maintainable with a heredoc, like:

```bash
curl -X POST -d - -H "Content-Type: application/json" $DEALBOT_CONTROLLER_ENDPOINT/tasks/retrieval <<EOF
{
  "Miner": $MINER,
  "Tag": $TAG,
  "PayloadCID": "$PAYLOAD",
  "CarExport": false,
  "Schedule": "$(date -v+5M +"%M %H * * *")",
  "ScheduleLimit": 1
}
EOF
```

* This kind of EOF does expansion like double quotes, so you can still do what you need
* Since the start/end tokens are the EOF word itself, you don't need to quote double quote characters, which is nice for JSON
* Lends itself much better to large strings, since you can split it into multiple lines

Note that I removed the `jq -c` here; as far as I can tell that's a no-op, as long as your expanded JSON here is valid. If it were invalid, I imagine there's no big difference between `jq` or `curl` erroring with a similar message.

mvdan (approved):

PR #225, opened 2021-06-16
don't publish deals that have failed due to local configuration

willscott:


controller/state/statedb.go
```diff
@@ -742,14 +742,37 @@ func (s *stateDB) PublishRecordsFrom(ctx context.Context, worker string) error {
...
+				return err
+			}
+			tsk := tskBuilder.Build().(tasks.FinishedTask)
+			if tsk.FieldErrorMessage().Exists() {
+				em := tsk.FieldErrorMessage().Must().String()
+				if strings.Contains(em, "there is an active retrieval deal with") ||
+					strings.Contains(em, "blockstore: block not found") ||
+					strings.Contains(em, "missing permission to invoke") ||
+					strings.Contains(em, "/efs/dealbot") ||
+					strings.Contains(em, "handler: websocket connection closed") {
```
mvdan, line 766: use a loop?

controller/state/statedb.go
```diff
@@ -742,14 +742,37 @@ func (s *stateDB) PublishRecordsFrom(ctx context.Context, worker string) error {
...
 				return err
 			}
-			c, err := cid.Decode(lnk)
+
+			// check if we should include it.
+			rcrdRdr, err := txContextLoader(ctx, tx)(cidlink.Link{c}, ipld.LinkContext{})
+			if err != nil {
+				return err
+			}
+			tskBuilder := tasks.Type.FinishedTask.NewBuilder()
```
mvdan, line 755: you might want the repr prototype :)

mvdan (approved):

controller/state/statedb.go
willscott: you mean put these strings in an array and loop through them?

controller/state/statedb.go
willscott: the non-repr version can be used for decoding, right? I've only run into the repr issue on the encoding side... have i done something wrong? 😱

controller/state/statedb.go
mvdan: Yes :)

controller/state/statedb.go
mvdan: As far as I'm aware, if you use the repr node for encoding, you should use the repr prototype for decoding. Otherwise it's a mismatch. A struct with stringjoin repr encoded as repr will result in a string, but decoded into the non-repr struct will fail, for example.

PR #231, opened 2021-06-18
Reset dealbot tasks when dealbot comes on line

hannahhoward:
# Goals

When a dealbot shuts down without draining (i.e. unexpected shut down / crash), reset its in progress tasks.

# Implementation

- State DB method and test to reset in progress tasks for a given dealbot to as if they've never been run
- Controller endpoint for the reset workers method
- Code in engine to call reset workers on startup

### DIFF
--- a/controller/state/statedb_dml.go
+++ b/controller/state/statedb_dml.go
@@ -97,4 +97,15 @@ const (
 	queryHeadSQL = `
 		SELECT cid FROM record_updates WHERE status = $1 AND worked_by = $2
 	`
+
+	workerTasksByStatusSQL = `
+	SELECT tasks.uuid, tasks.data FROM tasks
+	INNER JOIN task_status_ledger ON tasks.uuid=task_status_ledger.uuid
+	WHERE tasks.worked_by = $1 AND task_status_ledger.status = $2
+`
+
+	unassignTaskSQL = `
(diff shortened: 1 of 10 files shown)


controller/state/statedb_dml.go
```diff
@@ -97,4 +97,15 @@ const (
...
 	`
+
+	workerTasksByStatusSQL = `
+	SELECT tasks.uuid, tasks.data FROM tasks
+	INNER JOIN task_status_ledger ON tasks.uuid=task_status_ledger.uuid
+	WHERE tasks.worked_by = $1 AND task_status_ledger.status = $2
+`
+
+	unassignTaskSQL = `
+		UPDATE tasks SET data = $2, worked_by = NULL, cid = $3
```
willscott, line 108: should we do this, or should we make a new initial state task? having a log that jumps back to unassigned after getting partially executed may be harder to infer anything about when we look for source of failure at the end.

willscott (approved):
cc @gammazero to make sure this won't conflict with your schedule refactor too much

controller/state/statedb_dml.go
hannahhoward: I'm not entirely sure why I feel this way, but I feel that preserving the UUID will be relevant. Note this does actually wipe the logs that the reputation service will see.

PR #234, opened 2021-06-20
flesh out controller ux

willscott:
use esbuild over an npm package for frontend js

image


Dockerfile
```diff
@@ -6,8 +12,10 @@ RUN apk add --update gcc>=9.3.0 g++>=9.3.0 alpine-sdk
 WORKDIR /go/src/app/
 
 COPY . .
+COPY --from=js /usr/src/app ./controller/app
 # Fetch dependencies.
 RUN go get -d -v ./...
+RUN go generate ./...
```
mvdan, line 18: should this not be committed to master as-is? if we're worrying about forgetting to re-run go generate, maybe we should make that a CI check instead.

controller/app/.gitignore
```diff
@@ -0,0 +1,2 @@
+node_modules
```
mvdan, line 1: might as well put this in the root gitignore

controller/controller.go
```diff
@@ -1,3 +1,5 @@
+//go:generate go run github.com/filecoin-project/dealbot/controller/webutil/gen "./app" static/script.js
```
mvdan, line 1: a bit shorter:

> go run ./webutil/gen app static/script.js

controller/static/style.css
```diff
@@ -0,0 +1,51 @@
...
+
+    .cron-picker-dow button.active {
+        background: #0c63e4;
+    }
+  
+  .cron-picker-time select {
+      width: 60px;
+      display: inline-block;
+      margin: 0 5px;
+  }
```
mvdan, line 50: the indentation here seems to be off?

controller/webutil/gen/gen.go
```diff
@@ -0,0 +1,34 @@
...
+	cmd.Dir = os.Args[1]
+	err := cmd.Run()
+	if err != nil {
+		fmt.Printf("Failed to install frontend dependencies: %v\n", err)
+		if _, err := os.Stat(path.Join(os.Args[1], "node_modules")); os.IsNotExist(err) {
+			os.Exit(1)
+		}
+	}
+
+	data := webutil.Compile(os.Args[1], true)
```
mvdan, line 29: I wonder if it might be easier to just plop a `gen.go` file in the output directory that has:

```
//go:generate npm install
//go:generate go run github.com/evanw/esbuild/cmd/esbuild [whatever args]
```

This is assuming that what we're doing with esbuild could also be done via the command line. Since we're requiring npm to be installed anyway, I don't think calling it via os/exec gains us that much. I'll ultimately defer to you though, as I've never used esbuild.

controller/webutil/gen/gen.go
```diff
@@ -0,0 +1,34 @@
...
+	if len(os.Args) < 3 {
+		fmt.Printf("%s\n", data)
+	}
+	ioutil.WriteFile(os.Args[2], []byte(data), 0644)
```
mvdan, line 33: nit: error check

controller/app/index.js
```diff
@@ -0,0 +1,93 @@
...
+    })
+
+    if (!$('#newRepeat').is(':checked')) {
+        $("#setschedule").hide();
+    }
+    $("#newRepeat").on('change', () => {
+        if ($('#newRepeat').is(':checked')) {
+            $("#setschedule").show();
+        } else {
+            $("#setschedule").hide();
```
mvdan, line 30: isn't there a "toggle show/hide" method, or something like "set hidden as per this boolean"?

mvdan (commented):
The generate/docker/Go side SGTM.

The JS/HTML part looks fine too, though my experience there is more limited.

Dockerfile
mvdan: I worded my first sentence weird. I mean that the output of `go generate` should always be in master, so I don't think re-running go generate gets us anywhere.

Speaking of which, I don't get the `go get -d -v ./...` either. Modules will download dependencies as needed in the next step already.

willscott: I feel like `script.js` as a generated code that changes with every commit to frontend code is easier to build for docker rather than have in each commit as a way to not have to deal with merge conflicts

controller/webutil/gen/gen.go
willscott: This setup lets me run a development version of the controller that builds the frontend code from source every time it serves it, rather than using the pre-built version.

I can finish putting that in place so it's clear why i'm structuring it this way

controller/app/index.js
willscott: possibly?

Dockerfile
mvdan: Ah, I see. Maybe move it out of `go generate`, and into a build script of some sort, then? Because it is generally understood that the output of `go generate ./...` should be committed, so this will end up being confusing or problematic sooner or later. You can include its output dir in gitignore, though that adds to the confusion, if you ask me :)

controller/webutil/gen/gen.go
mvdan: Ah I see, makes sense.

Dockerfile
willscott: the output is in gitignore :)

controller/app/index.js
```diff
@@ -0,0 +1,94 @@
...
+
+        if ($('#newRepeat').is(':checked')) {
+            data.Schedule = $('#newSchedule').val()
+            data.ScheduleLimit = $('#newScheduleLimit').val()
+        }
+        if ($('#newScheduleTag').val() !='') {
+            data.Tag = $('#newScheduleTag').val()
+        }
+
+        $.ajax({
```
hannahhoward, line 85: I wonder: for $.ajax, maybe it isn't worth the jquery? $.ajax was just a thing that made things easier before `fetch` existed. Fetch returns a promise and then you can just do Promise.All on the an array of the promises and avoid the weird done/remaining log.

Non-blocking.

controller/state/statedb.go
```diff
@@ -752,7 +752,7 @@ func (s *stateDB) PublishRecordsFrom(ctx context.Context, worker string) error {
 			if err != nil {
 				return err
 			}
-			tskBuilder := tasks.Type.FinishedTask.NewBuilder()
+			tskBuilder := tasks.Type.FinishedTask__Repr.NewBuilder()
```
hannahhoward, line 755: This seems quite unrelated? What does it fix?

controller/controller.go
```diff
@@ -121,7 +121,7 @@ func (fw *logEcapsulator) Write(p []byte) (n int, err error) {
 //go:embed static
 var static embed.FS
 
-func NewWithDependencies(listener, graphqlListener net.Listener, gqlToken string, recorder metrics.MetricsRecorder, backend state.State) (*Controller, error) {
+func NewWithDependencies(ctx *cli.Context, listener, graphqlListener net.Listener, recorder metrics.MetricsRecorder, backend state.State) (*Controller, error) {
```
hannahhoward, line 124: Nit: the purpose of seperating New from NewWithDependencies was to put the CLI parsing in a single place, so that tests wouldn't worry about assembling a CLI. Technically, NewWithDepedencies could be a private function and the tests could skip the _test underscore in the package name

hannahhoward (commented):
Left comments -- mostly non-blocking but I would like to understand what the change to state.go is about, since that seems unrelated.

controller/state/statedb.go
willscott: I think it was either @mvdan or @rvagg  who mentioned that we should be de-serializing into __Repr's as well as serializing into them. This is a remnant of the constant 'the default schema'd types are not the ones you actually should be using problem in the current ipld-prime' 
in practice - it shouldn't matter any any valid representation _should_ de-serialize into the nonrepresentation type, but i think that might not be guaranteed?

hannahhoward (approved):

controller/state/statedb.go
mvdan: This was me, yes. In practice, in this particular case, it might not matter. But it's certainly asking for trouble. This would definitely break if we added a struct field with non-default representation, for example.

PR #244, opened 2021-06-23
Move task scheduler into controller

### LINKED ISSUE #227: move scheduler to run in controller rather than in daemon
currently a daemon commits to do all iterations of a schedule task. we ideally have that scheduling happen in controller to: 

* allow daemons to rotate without losing subsequent iterations
* allows canceling subsequent iterations of a repeated task

gammazero:
### Controller Schedules Tasks
A scheduled task is not run directly but is used to schedule the creation of runnable (non-scheduled) tasks.  When a new task is given to the controller, if the task has a schedule, then that task is assigned to the scheduler and a task generation job is scheduled.  This task will not be available to PopTask since it owned by the scheduler.  When the task generation job runs, a new runnable task is written to the database.  It has no schedule, no owner, and a new UUID; it can be picked up by PopTask.  

### Task Recovery
If the controller is restarted, all tasks that were previously scheduled are recovered.  This is done by reading in all tasks that are assigned to the scheduler and scheduling them

### Changing a Task Schedule or Removing a Scheduled Task
If a change is made to a task's schedule, the change will be detected the next time the task is run and the task will be rescheduled to run on the new schedule.  If a scheduled task is removed, this is detected and the scheduled job for the task is stopped.

### Schedule Expiration
If a scheduled task has run for the full duration of its schedule limit, then the scheduled job for the taks is stopped and the scheduled task is marked as expired.  This prevents expired tasks from being recovered.  An expired task can be restarted by removing the task's ownership in the task DB table. 

### Adding Scheduled Tasks Out-of-band
If a scheduled task is added directly to the database, outside of the normal task ingestion path, the task will be detected and scheduled.

This fixes #227


controller/state/statedb.go
```diff
@@ -328,20 +360,66 @@ func (s *stateDB) AssignTask(ctx context.Context, req tasks.PopTask) (tasks.Task
...
+			// No more tasks to pop
+			return nil, nil
+		}
+		if !scheduled {
+			// Found a task that is runable, stop looking for tasks
+			assigned = task
+			break
+		}
+		// Got a scheduled task, so schedule it.
+		err = s.scheduleTask(task, nil)
```
hannahhoward, line 381: This seems like a really unusual time to schedule a task -- it requires a daemon to attempt to pop it? What if daemons are busy and you miss the first schedule? Why not put scheduling when you initially create the task?

hannahhoward (changes requested):
I feel pretty strongly that initial scheduling should happen at the point of injestion -- when NewStorageTask or NewRetrieval task is called. This would obviate the need for a loop in PopTask, I think, as the scheduled tasks would already be assigned.

controller/state/statedb.go
gammazero: Was thinking, why schedule tasks if no one is asking for them.  If we want to start up a controller and load a bunch of tasks, but then sometime later start daemons, this may be nice.  I suppose that whether it is better to worse to miss the first schedule depends on the type of schedule (run every n hours, or run at a specific time).

So, will move scheduling to point of ingestion as that is probably least surprising.

gammazero: Tasks are now scheduled on ingestion.  However, I left the ability to detect and schedule unassigned scheduled tasks here so that if we want to insert a scheduled task into the DB outside of the normal ingestion path, it will still get picked up.  Otherwise, this the code here should never see a scheduled task.

hannahhoward: I would say we specifically want things coming in through the normal ingestion path -- NewStorageTask and NewRetrievalTask, and nothing else. I'm trying to think what else could happen here -- inserting through PG Admin? I'd hope we never do that.

But I don't feel so strongly I'd block the PR on it.

controller/state/statedb.go
```diff
@@ -124,7 +127,7 @@ func migrateDatabase(dbName string, dbInstance database.Driver) error {
 }
 
 // NewStateDB creates a state instance with a given driver and identity
-func NewStateDB(ctx context.Context, driver, conn string, identity crypto.PrivKey, recorder metrics.MetricsRecorder) (State, error) {
+func NewStateDB(ctx context.Context, driver, conn string, identity crypto.PrivKey, recorder metrics.MetricsRecorder, runNotice chan string) (State, error) {
```
hannahhoward, line 130: Nit: when adding parameters that are only used in test, I prefer having a second constructor function that has the additional params, and having the first constructor call the second with a nil value -- that way you don't have the extra param in regular non-test code (see controller/controller.go).

This is assuming you don't see a use for run notice outside of test.

controller/state/taskscheduler.go
```diff
@@ -0,0 +1,252 @@
...
+		}
+	}
+	return false
+}
+
+func getTaskSchedule(task tasks.Task) (string, time.Duration) {
+	var schedule, limit string
+	var duration time.Duration
+
+	if t := task.RetrievalTask; t.Exists() {
```
hannahhoward, line 225: goodness this is a sad state of affairs or go-ipld-prime usability :P (not your fault)

btw, you could make this easier by righting a TaskSchedule method over in the tasks module on the Task itself, and then you could use the private members to make it a bit simpler.

hannahhoward (approved):
Remaining issues are non-blocking

controller/state/statedb.go
gammazero: If `popTask` finds an unassigned task that has a schedule, there are two things that can be done:
1. Schedule the task (what is done now)
2. Log error and ignore task

With the way it is now (option 1), the new dealbot will pick up any unassigned scheduled tasks that the previous dealbot may not have popped yet.  With option 2, any of those tasks would have to be removed and re-inserted.

controller/state/statedb.go
gammazero: Yeah, I was considering that this could be useful functionality... but since I do not have a specific use for it yet, probably better to hide it behind a test-only constructor.

controller/state/taskscheduler.go
gammazero: I like that idea better, will do.

controller/state/statedb.go
gammazero: Logging a warning when a scheduled task with no owner is seen.

PR #249, opened 2021-06-24
support separate file to output finalized tasks to

willscott:


raulk (commented):
Can we write both ongoing events and finalized tasks, and identify each through a `type` field with values `ongoing_event`, `completed_task`?

This would allow us to track live dealbot activity and final data through ES/Kibana.

willscott:
updates so the output struct is updated on stage transitions both for in progress and finalized tasks.

you can split on the top-level `status` field in the output json log lines, which will be `1` or `2` for `ongoing_event` and `3`(success) or `4`(failure) for `completed_task`'s

raulk (approved):
Can't fully say if the code is right, but if it does what it says on the tin, it should be in.

controller/state/statedb.go
```diff
@@ -591,6 +615,29 @@ func (s *stateDB) Update(ctx context.Context, taskID string, req tasks.UpdateTas
 	return task, nil
 }
 
+func (s *stateDB) log(ctx context.Context, task tasks.Task, tx *sql.Tx) {
+	finalized, err := task.Finalize(ctx, txContextStorer(ctx, tx))
```
hannahhoward, line 619: this is going to error if the task is not Successful / errored -- see https://github.com/filecoin-project/dealbot/blob/main/tasks/model.go#L348

is this what you want? It looks like you're trying to log all updates, but I think this is going to cause your code to skip updates that aren't to a success state.

hannahhoward (changes requested):
Eek I think I see an issue but it's already merged -- see comment

PR #252, opened 2021-06-25
Skip task generation if previous task incomplete

### LINKED ISSUE #251: Do not generate new task from schedule unless previous task is finished
There are 2 deficiencies with the current implementation of scheduled task generation:
1. The generated task does not have any information that identifies its parent
2. A new task is generated even if the previously generated task has not yet been finished.

Fixing 2 requires fixing 1.

gammazero:
Migration creates a new parent column in the tasks table. If a task is generated from a scheduled task, then the generated task's parent is set to the ID of the scheduled task. When a scheduled task is run, first a query looks for any tasks that are children of the scheduled (generator) task, and if any are unfinished (status < Successful), then task generation is skipped.

This fixes #251


willscott (approved):

hannahhoward (approved):
LGTM

PR #254, opened 2021-06-25
variant of finalize for logging

willscott:


tasks/locallink.go
```diff
@@ -0,0 +1,26 @@
...
+	ipld "github.com/ipld/go-ipld-prime"
+)
+
+type localLink struct {
+	ipld.Node
+}
+
+var _ ipld.Link = (*localLink)(nil)
+
+func (l *localLink) LinkBuilder() ipld.LinkBuilder {
```
hannahhoward, line 16: Just a heads up this is all going to need to be rewritten if we switch to LinkSystem

hannahhoward (approved):
Approving, but wondering why we don't just write the normal node with a different event per @raulk's suggestion here? It seems odd to try to finalize a node that's not really final.

tasks/locallink.go
willscott: hmm. this actually won't work because of https://github.com/ipld/go-ipld-prime/blob/master/codec/dagjson/marshal.go#L191 :-|

willscott: ok. i ended up making my own codec for the purposes of json-ing in the way I want and got thwarted again by the great ipld 0.7/0.9 fissure, but i think this logs in the way i want now :)

PR #255, opened 2021-06-26
flag /secret passing so that ajax can be credentialed

willscott:
env var specifying "username:password" will turn into a jsonp fetch (which will get implicit creds) and will be kept in local browser state so that subsequent ajax can also be credentialed


hannahhoward (approved):
LGTM

I think this is ok from a security standpoint. I think. We can certainly revisit if we get to publishing to rep services.

PR #256, opened 2021-06-26
allow a retrieval schedule to be defined on storage tasks

### LINKED ISSUE #215: Only attempt retrieval on successfully stored data
This issue is to track work in progress to ensure that a full storage+retrieval deal cycle only tests retrieval if the storage deal succeeded.  There's no point in testing retrieval of CIDs that we already know were not successfully stored, since this will always fail.

willscott:
when storage succeeds, retrievals then get scheduled on that schedule

fix #215

### DIFF
--- a/controller/app/package-lock.json
+++ b/controller/app/package-lock.json
@@ -1,8 +1,44 @@
 {
   "name": "dealbot-controller",
   "version": "1.0.0",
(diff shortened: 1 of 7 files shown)


hannahhoward (approved):
LGTM by me! Cool!

PR #260, opened 2021-06-29
Delete tasks endpoint

hannahhoward:
# Goals

Be able to cancel either task schedules (not the running instances of scheduled tasks) or unstarted tasks

# Implementation

This turned out to be enjoyable difficult, as SQLite apparently does not implement ALTER TABLE ADD/DROP CONSTRAINT. This led after much discussion to the idea to drop SQLite.

Here is the approach used to defaulting to PG:
- most of us are never running integration tests locally any more, but if we do, everyone can handle their own setup
- what we are doing frequently is writing unit tests
- unit tests are an interesting problem cause not only do they need a database but it needs to be reset completely between tests
- previously, we did this my just making a new temp directory and SQLite file
- now, upon test start, we check if we can connect to a running PG instance on a host machine (if the dev is running PG locally, they can set env vars so that it connects)
- if there is no host  machine instance, start a temporary docker container for postgres. Connect there and tear it down on test terminate.
- for resetting, since it's very time consuming to actually try to tear down and restart docker instances, we instead just reset all records in the db using truncate, a pretty standard practice for resetting databases for unit tests (that talk to the DB of course)

Otherwise, the implementation is pretty straightforward.
- SQL queries
- State method
- Controller endpoint
- Controller client method


willscott:
lets get rid of the sqlite support rather than adding this much complexity to support it

controller/state/statedb.go
```diff
@@ -1078,3 +1083,29 @@ func (s *stateDB) ResetWorkerTasks(ctx context.Context, worker string) error {
 	}
 	return nil
 }
+
+func (s *stateDB) Delete(ctx context.Context, uuid string) error {
+	task, _, err := s.getWithTag(ctx, uuid)
```
gammazero, line 1088: It might be better to create a version of this function that is not transaction-wrapped and call to get and delete within the same transaction.

controller/state/statedb_dml.go
```diff
@@ -132,4 +132,12 @@ const (
 		UPDATE tasks SET worked_by = $2
 		WHERE uuid = $1
 	`
+
+	deleteTaskSQL = `
+		DELETE FROM tasks WHERE uuid = $1
+	`
+
+	deleteTaskStatusLedgerSQL = `
+		DELETE FROM task_status_ledger WHERE uuid = $1
```
gammazero, line 141: It might make sense to combine these into a single query:
```
deleteTaskAndStatueSQL = `
    DELETE FROM tasks WHERE uuid = $1;
    DELETE FROM task_status_ledger WHERE uuid = $1
`
```

gammazero (commented):
A few suggestions, but looks good.

controller/state/statedb.go
hannahhoward: agree.

transaction wrapping should happen in the calling code for a utility function like this, IMHO.

controller/state/statedb_dml.go
hannahhoward: oh did I not push the fix? you actually can great ride of this with ON DELETE CASCADE

willscott (approved):

README.md
```diff
@@ -49,8 +49,8 @@ Dealbot Controller
 | graphql | `DEALBOT_GRAPHQL_LISTEN` | exposed `host:port` for external public graphql queries |
 | metrics | `DEALBOT_METRICS` | either `prometheus` to expose a `/metrics` api, or `log` to write metrics to stdout |
 | identity | `DEALBOT_IDENTITY_KEYPAIR` | filepath of a libp2p identity to sign public records of dealbot activity |
-| driver | `DEALBOT_PERSISTENCE_DRIVER` | `sqlite` or `postgres` |
-| dbloc | `DEALBOT_PERSISTENCE_CONN` | the file (for sqlite) or db conn string from postgres |
+| driver | `DEALBOT_PERSISTENCE_DRIVER` | `postgres` |
```
willscott, line 52: get rid of entirely?

hannahhoward: I thought that was what we decided?

willscott: but now that there's only one entry we can get rid of the flag entirely?

PR #262, opened 2021-07-01
render data-table

willscott:
screenshot:

image


hannahhoward (approved):
LGTM but... we may need a follow-on PR to make the table collabsible -- I think it's going to be quite large.

PR #265, opened 2021-07-01
Remove any extra suffix from basic auth information

hannahhoward:
# Goals

Strip newlines at end of basicauth value as a precaution, since it produces a javascript error


willscott (approved):
🙃

PR #268, opened 2021-07-02
fix various logging errors

coryschwartz:
Currently the daemon logs include such lines as this:

```
{"level":"dpanic","ts":"2021-07-01T16:20:55.361Z","logger":"dealbot","caller":"zap@v1.16.0/sugar.go:179","msg":"Ignored key without a value.","ignored":"XXX"}
```

which is caused by not using a key on `Infow`. Found a couple more
instances of that.

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -260,13 +260,13 @@ func (e *Engine) tryPopTask() tasks.Task {
 		return nil
 	}
 
-	log.Infow("successfully acquired task", "uuid", task.UUID.String())
+	log.Infow("successfully acquired task", "taskID", task.UUID.String())
 	return task // found a runable task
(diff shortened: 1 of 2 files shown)


willscott (approved):

engine/engine.go
```diff
@@ -260,13 +260,13 @@ func (e *Engine) tryPopTask() tasks.Task {
 		return nil
 	}
 
-	log.Infow("successfully acquired task", "uuid", task.UUID.String())
+	log.Infow("successfully acquired task", "taskID", task.UUID.String())
```
coryschwartz, line 263: This change was because "taskID" is the key used elsewhere for task uuid.

PR #273, opened 2021-07-05
allow access to previous record updates from graphql

willscott:


controller/state/statedb_test.go
```diff
@@ -603,7 +603,7 @@ func TestComplete(t *testing.T) {
 		require.NoError(t, state.DrainWorker(ctx, "dealbot1"))
 		require.NoError(t, state.PublishRecordsFrom(ctx, "dealbot1"))
 
-		nextHead, err := state.GetHead(ctx)
+		nextHead, err := state.GetHead(ctx, 0)
```
hannahhoward, line 606: nb: we should add a test of getting a previous head to insure it works.

hannahhoward (commented):
LGTM

Issue #271, opened 2021-07-04
Expose Deal Proposal CID through GraphQL endpoint


raulk:

hannahhoward:
@raulk for retrievals, I assume you don't care about Proposal CID, but rather retrieval DealID -- that's what is tracked in the FSM for retrievals I think.

hannahhoward:
Also do you need on chain deal ID for Storage or is ProposalCID sufficient?

hannahhoward:
@willscott 's PR above attempts to get more data but doesn't actually work. I want to confirm the minimum viable set of things you need, which based on my existing knowledge of markets software is ProposalCID for Storage Deals and Retrieval DealID for retrieval deals (which based on my knowledge of markets software, unless it's changed, are how the FSMs are key'd)

raulk:
Thanks for checking. Let's go with your call @hannahhoward. For storage deals I think the ProposalCID would be enough as that's what enables us to correlate with the journal, but the DealID would be good to have, as you noted. For retrievals, I agree with you.

willscott:
these are now available

PR #274, opened 2021-07-05
Logging of dealid and proposalID during deals, and reporting in finalized task

willscott:


tasks/retrieval_deal.go
```diff
@@ -219,7 +219,35 @@ func (de *retrievalDealExecutor) executeAndMonitorDeal(ctx context.Context, upda
 				if event.Event == retrievalmarket.ClientEventDealAccepted {
 					stage = RetrievalStageDealAccepted
 					dealStage = RetrievalStages[stage]()
-					err := updateStage(ctx, stage, dealStage)
+					//there should be a deal now.
+					allRetrievals, err := de.node.ClientListRetrievals(ctx)
```
willscott, line 223: are there other things to match on here in pulling out the deal from the retrieval? cc @hannahhoward

willscott:
fix #271

tasks/retrieval_deal.go
```diff
@@ -219,7 +219,35 @@ func (de *retrievalDealExecutor) executeAndMonitorDeal(ctx context.Context, upda
...
+								break
+							}
+						}
+					}
+					if expectedID != 0 {
+						allDeals, err := de.node.ClientListDeals(ctx)
+						if err != nil {
+							return err
+						}
+						for _, maybeOurDeal := range allDeals {
```
hannahhoward, line 244: This isn't going to work -- these are not the same DealID. The DealID on storage side is an onchain incrementing integer for all deals in the network. The DealID for retrieval is different -- it's just an off chain incrementing integer per client that has nothing to do with the storage deal id.

tasks/retrieval_deal.go
```diff
@@ -219,7 +219,35 @@ func (de *retrievalDealExecutor) executeAndMonitorDeal(ctx context.Context, upda
...
+					//there should be a deal now.
+					allRetrievals, err := de.node.ClientListRetrievals(ctx)
+					expectedPayload, parseErr := cid.Parse(de.task.PayloadCID)
+					if parseErr != nil {
+						return err
+					}
+					expectedID := abi.DealID(0)
+					if err == nil {
+						for _, maybeOurRetrieval := range allRetrievals {
+							if maybeOurRetrieval.PayloadCID.Equals(expectedPayload) {
```
hannahhoward, line 231: Miner ID Would be another thing we can check -- that's what we use above.

tasks/retrieval_deal.go
hannahhoward: There's not a really great correlation here -- but we could use payload CID + miner ID.... except that can definitely be duplicated. Our best piece of information is what we have locally when we create the retrieval from the storage deal -- when we create the retrieval deal after a storage, we should log the proposal CID.

tasks/storage_deal.go
```diff
@@ -379,7 +384,14 @@ func (de *storageDealExecutor) executeAndMonitorDeal(ctx context.Context, update
 			}
 
 			if len(info.DealStages.Stages) > 0 {
-				err = updateStage(ctx, storagemarket.DealStates[info.State], toStageDetails(info.DealStages.Stages[len(info.DealStages.Stages)-1]))
+				newState := storagemarket.DealStates[info.State]
```
hannahhoward, line 387: This also it not going to work. We don't get the DealID till the deal is on chain unfortunately, so we have to wait till at least the publish point to get the deal ID

hannahhoward (changes requested):
Most of this code is not actually going to work. I'll reach out 1-on-1 to see if we can get the best approach here.

hannahhoward (approved):

PR #283, opened 2021-07-09
use helm charts that supports k8s driver

ognots:
this helm charts provides the kubeconfig necessary to properly enable self-serve deployments


hannahhoward (approved):
LGTM

PR #293, opened 2021-07-14
Count number of cids in retrieved deals

willscott:


tasks/retrieval_deal.go
```diff
@@ -268,6 +270,30 @@ func (de *retrievalDealExecutor) executeAndMonitorDeal(ctx context.Context, upda
 					return err
 				}
 
+				// See if car and estimate number of CIDs in the piece.
+				fp, err := os.Open(filepath.Join(de.config.DataDir, de.task.PayloadCID.x))
```
hannahhoward, line 274: if wonder if we should just look at `ref.IsCar`  or `de.task.CARExport.x` (which is what `ref.IsCar` is set from) instead of just attempting a file open.

One reason: if the retrieval is a single, non-car file, this part will success and then log CARReadClean: false, which actually isn't an error.

hannahhoward (approved):
One suggestion, otherwise LGTM

tasks/retrieval_deal.go
willscott: we should be able to know that from the retrieval task that's also logged. I think it's interesting to still try and understand how many of the sectors actually are cars. (for instance, the single file stored as a deal could actually be a valid car file still, right?)

PR #306, opened 2021-07-19
switch shared storage back to efs

ognots:
shared storage must be read-writeable by all.
emptyDir volumes currently provide insufficient space for dealbot daemons which need to stage 32GB sectors

### DIFF
--- a/controller/spawn/base.yaml
+++ b/controller/spawn/base.yaml
@@ -65,6 +65,12 @@ application:
         name: dealbot
   ingress:
     enabled: false
+  storage:
+    - mount: /shared
+      volume:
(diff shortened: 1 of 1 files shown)


coryschwartz (approved):

hannahhoward (approved):
LGTM from my end,  if you guys have finished working out changes

coryschwartz:
@ognots  They don't; they expect the directory to already exist. However, the file names they use are hashes.

ognots:
so long as all dealbot daemons are able to account for their own directories/cids let's ship this to unblock retrieval. we can look at improving tree structure later

PR #308, opened 2021-07-19
Remove all on retrieval cleanup!

hannahhoward:
# Goals

Retrievals of directories are not getting cleanup properly, as os.Remove only deletes empty directories.

# Implementation

os.RemoveAll is needed to delete non empty directories


ognots (approved):
LGTM

PR #309, opened 2021-07-19
Remove automatic daemon deployment

hannahhoward:
# Goals

Move to all self-serve daemons

# Implementation

No longer deploy new daemons with releases

(we can manually shut down the running instances and then they won't get restarted on deploy)


coryschwartz (approved):

coryschwartz:
Glad to see these go away.

PR #331, opened 2021-09-30
use circleci context for docker login

coryschwartz:


coryschwartz:
previously these were env vars at the project level, but there is a circle context we can use instead.

willscott (approved):

PR #346, opened 2021-11-28
use port list

coryschwartz:
fixes: https://github.com/filecoin-project/dealbot/pull/new/port-list

Use a list of ports rather than a single port.

This was a change in the chart.

Also, this error points out another bug -- the `Protocol` (capital P) should be `protocol` in the defualt, but that is another matter. This fix overrides the default so the dealbot's own ports are used in the chart.


willscott (approved):

coryschwartz:
Needed to include the namespace for install as well. This is required so the chart can read the namespace with `Release.Namespace`. Tested, and it works.

PR #359, opened 2022-01-04
publish results over legs

willscott:
* [x] waiting for [go-ds-sql](https://github.com/ipfs/go-ds-sql/pull/25) datastore compatibility for use as the datastore behind the legs storage
* [x] the libp2p host used for legs should have a configurable listening address, so that inbound connections can be received
* [x] Update use of lotus retrieval API to work with version of lotus that supports context data stores


tasks/retrieval_deal.go
```diff
@@ -107,23 +108,37 @@ func (de *retrievalDealExecutor) queryOffer(logg logFunc) error {
...
+	qo, err := de.node.ClientMinerQueryOffer(de.ctx, de.minerAddress, payloadCid, nil)
 	if err != nil {
 		return err
 	}
+	ro := qo.Order(de.config.WalletAddress)
+	de.offer = v0api.RetrievalOrder{
+		Root:                    ro.Root,
+		Piece:                   ro.Piece,
+		Size:                    ro.Size,
+		FromLocalCAR:            "",
```
masih, line 120: Zero value.
```suggestion
```

controller/state/statedb.go
```diff
@@ -144,6 +148,19 @@ func newStateDBWithNotify(ctx context.Context, dbConn DBConnector, migrator Migr
 		log.Infow("recovered scheduled tasks", "task_count", count)
 	}
 
+	storeLS := storeLS(st.Store(context.Background()))
```
hannahhoward, line 151: part of me would suggest exposing the datastore here, but having instantiation of libp2p and legs be elsewhere. seems like an odd mixing of concerns to setup a whole network transport stack inside of a database instantiation.

hannahhoward: I could be wrong but I don't think I actually see any real common deps other than the priv key?

hannahhoward (approved):
LGTM, but I personally would make the suggested change around setting up legs

controller/state/statedb.go
```diff
@@ -144,6 +148,19 @@ func newStateDBWithNotify(ctx context.Context, dbConn DBConnector, migrator Migr
...
+	host, err := NewHost(identity, addrs)
+	if err != nil {
+		return nil, err
+	}
+
+	b := dbDS("legs_data", st.db())
+	pub, err := dtsync.NewPublisher(host, b, storeLS, "/dealbot/v1.0.0")
```
masih, line 158: Can we make the topic configurable? 

Side note: should this be Pando topic or is the expectation that Pando producers can also specify topic?

masih (approved):

controller/state/statedb.go
willscott: happy to use a pando topic if there is one

controller/state/statedb.go
willscott: it's a good point. i just don't know where i actually want to make it the legs/libp2p host :)

PR #366, opened 2022-02-16
Add option to specify bootstrap addresses

masih:
Allow the libp2p host to optionally perform bootstrap with given list of
addinfos.


controller/state/libp2p.go
```diff
@@ -87,6 +89,15 @@ func NewHost(priv crypto.PrivKey, listenAddrs []multiaddr.Multiaddr) (host.Host,
 	return host, nil
 }
 
+func bootstrapHost(host host.Host, btstrp []peer.AddrInfo) (io.Closer, error) {
+	bCfg := bootstrap.DefaultBootstrapConfig
+	bCfg.MinPeerThreshold = 1
```
willscott, line 94: should this be the number of items in the btstrp?

willscott (approved):

controller/state/libp2p.go
masih: TLDR; not sure.

That determines how many bootstrap connections to open. In the case of Pando integration it will most likely remain to 1. Though I am not sure what the future brings. Setting that to the length of addrinfos when a lot of addresses are provided might have unintended consequences. 

The default value for this is 4. I'll leave a todo to parameterise it since it is entirely application specific.

PR #371, opened 2022-02-23
Gitops readiness

coryschwartz:
This is being deployed in the us-east-2-dev cluster so we can test gitops deployment

### DIFF
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -1,0 +1,5 @@
+# deployment manifests
+
+## documentation:
+
+https://fluxcd.io/
--- a/deploy/base/kustomization.yaml
+++ b/deploy/base/kustomization.yaml
@@ -1,0 +1,6 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - release.yaml
+  - source.yaml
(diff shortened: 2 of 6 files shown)


coryschwartz:
To prepare for a future switch to weave, I've got a proof-of-concept "weave works" flux installed; I've set it up to watch this branch (gtitops-readiness) and deploy into a test namespace in us-east-2-dev. 

We don't have to switch to gitops deployment now (although we can if you want to try it out).

coryschwartz:
A single flux instance is running --this is intended to represent weave in the future--, It's s setup with permission to run deployments into our existing clsuters.


In a gitops way, you'd be able to control when something is installed into each cluster. Currently, you'll see there's only deploy/us-east-2-dev, but in the future, this would be deploy/us-east-1 or deploy/whatever-cluster

deploy/us-east-2-dev/secrets/sealed-secrets.yaml
```diff
@@ -0,0 +1,83 @@
+---
```
coryschwartz, line 1: These are "sealed secrets"

https://github.com/bitnami-labs/sealed-secrets

Each cluster you'd deploy into has a sealed-secret controller running, which have their own private/public keypair. 

These secrets remain encrypted until they are deployed into the destination cluster. From there, the sealed-secret controller running in the destination cluster will decrypt secrets.

This does make it a bit of a pain, yes. But it lets you store the secrets in git.

PR #374, opened 2022-03-01
Add option to specify external addresses advertised by legs

masih:
Add an option to explicitly set the external addresses advertised as
part of legs messages on gossipsub.

Upgrade legs to fix the bug in supplying DNS addresses.


masih:
Infra work here: https://github.com/filecoin-project/lotus-infra/pull/584

willscott (approved):

PR #378, opened 2022-03-02
Move retrieval to v1 api

hannahhoward:
# Goals

Get off deprecated APIs

# Implementations

Basically, this just takes lotus's retrieval CLI command's code and mashes it up with our existing retrieval code to hopefully produce the same set of updates back to the controller. We don't really have tests for retrieval, so I guess we'll see what it produces.

Hopefully the integration tests run in CI will give us a decent indicator of whether this has successfully done the thing?

### DIFF
--- a/lotus/lotus.go
+++ b/lotus/lotus.go
@@ -112,7 +112,7 @@ func NewAPIOpener(ctx *cli.Context) (*APIOpener, APICloser, error) {
 
 	apiInfo := cliutil.ParseApiInfo(tokenMaddr)
 
-	addr, err := apiInfo.DialArgs("v0")
+	addr, err := apiInfo.DialArgs("v1")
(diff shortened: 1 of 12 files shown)


lotus/lotus.go
```diff
@@ -60,7 +60,7 @@ func NewAPIOpenerFromCLI(cctx *cli.Context) (*APIOpener, APICloser, error) {
 		return nil, nil, fmt.Errorf("cannot connect to lotus api: missing --api or --repo flags")
 	}
 
-	addr, err := apiInfo.DialArgs("v0")
+	addr, err := apiInfo.DialArgs("v1")
```
willscott, line 63: change the dial args from v0 to v1 lower in this file as well

willscott (approved):

PR #423, opened 2022-05-25
Feat/ipld schema

sssion:
upgrade ipldschema from go gen to bindnode


willscott (commented):
Awesome, thanks for taking on this refactor!

Are you able to run it, accept and complete a task from a worker demon (you don't have to actually run a worker, just curl the endpoints to pop off a task, and then update it's status with a post), and finalize the demon so that the task becomes visible in the finishedtasks endpoint?

controller/graphql/schema.go
```diff
@@ -1,253 +1,36 @@
-// Code generated by github.com/ipld/go-ipld-prime/schema/gen/graphql/server, DO NOT EDIT.
```
willscott, line 1: are you no longer letting code-gen generate this?

sssion: because the generated schema is based on .schema file and the generated go files. But bindnode need not the generated go files

willscott: but this file is generating code using https://github.com/ipld/go-ipld-graphql - the generated is automatically created for graphql stubs based on the ipld schema.

you should be able to pull out a type system from the bind-node schema and use that to generate this file, or you can remove it fully.
hand-coding it is probably not great for maintainability

sssion: ok, I will remove it

